PR #5, opened 2014-07-29
Split Client into Client and LocalClient

Adam-:


syzop:
I've just done this myself now, so this can be closed.

PR #13, opened 2014-10-02
Update m_names to remove the trailing space

Clam-:
Removes the trailing space generated in RPL_NAMEREPLY responses, override last element to null before sending.


tmcarthur:
Sorry - it comes from line 222 - so that's probably where you could correct it.  Just be sure it doesn't have any negative side effects (and update pull request and I'll pull in ;) )

syzop:
pull req not updated. can file a new one if you really care ;)

PR #15, opened 2015-03-10
Fix build with LibreSSL

Sp1l:
LibreSSL has removed RAND_egd completely, this patch adds RAND_egd
detection to configure and adds HAVE_RAND_EGD checks to the code where
required.

Please consider removing EGD support completely, there's no OS that does not have a proper source of random for many years.

### DIFF
--- a/configure
+++ b/configure
@@ -3868,6 +3868,53 @@ fi
 
 
 
+ 
+ 
+ 
+{ $as_echo "$as_me:${as_lineno-$LINENO}: checking for RAND_egd in -lcrypto" >&5
+$as_echo_n "checking for RAND_egd in -lcrypto... " >&6; }
+if ${ac_cv_lib_crypto_RAND_egd+:} false; then :
+  $as_echo_n "(cached) " >&6
+else
+  ac_check_lib_save_LIBS=$LIBS
+LIBS="-lcrypto  $LIBS"
+cat confdefs.h - <<_ACEOF >conftest.$ac_ext
+/* end confdefs.h.  */
+
+/* Override any GCC internal prototype to avoid an error.
+   Use char because int might match the return type of a GCC
+   builtin and then its argument prototype would still apply.  */
+#ifdef __cplusplus
+extern "C"
+#endif
+char RAND_egd ();
+int
+main ()
+{
+return RAND_egd ();
+  ;
+  return 0;
+}
+_ACEOF
+if ac_fn_c_try_link "$LINENO"; then :
+  ac_cv_lib_crypto_RAND_egd=yes
+else
+  ac_cv_lib_crypto_RAND_egd=no
+fi
+rm -f core conftest.err conftest.$ac_objext \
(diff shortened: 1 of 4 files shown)


tmcarthur:
Thanks for the patch. I will have a look today but first glance looks good. And yeah we should look at removing - modern OS manages entropy better than OpenSSL a can hope to.

Sp1l:
It isn't even an OpenSSL thing... EGD is the python Entropy Gathering Daemon, all modern OSes have a proper source of randomness, even Windows. OpenSSL just supports using EGD as a source. Anyone requiring EGD will be running on a platform so old that it's highly unlikely they'll be able to run any recent software.

tmcarthur:
Perhaps, but compatibility is still a concern, so it's something for us to look into to verify there's not a reason to keep the support there and discourage it over removing it entirely.

In any event - thanks for the patch!

kaniini:
Compatibility with what exactly?

tmcarthur:
It's a possibility - I don't think it's likely, but I also don't mind including the ability to use EGD in these specific cases since it doesn't actually harm users that don't use EGD.

feld:
Keep in mind that we're talking about *very old commercial unix* here.

EGD needed until        OS release date
IRIX	6.5.19			feb 2003
Solaris 2.6                     jul 1997
AIX     5.2                     oct 2002
Tru64   5.1B                    sep 2002
HP-UX   11i v2                  sep 2003

tmcarthur:
I understand that - and it's probably fine to remove this support - but my bias is towards making sure I don't rip out the carpet from people, so I'd rather take some steps to assure that before removal, and then consider the removal.

feld:
certainly! :)

kaniini:
I encourage you to attempt building even UnrealIRCd 3.2 on these platforms (hint: I did it for you, it doesn't build on IRIX, AIX or Tru64.  I don't have access to the others.)

tmcarthur:
Okay, thanks for looking at that. As I noted I want to perform due diligence before removing this - this is part but not all of that.

Thanks for the information though it makes looking into this quite a bit easier!

kaniini:
As I recall, SunOS `dlopen()` is broken and is unsuitable for use in UnrealIRCd as well due to failing to properly handle dependency resolution.  So that removes SunOS from the list.

So, basically:

* IRIX, AIX, Tru64 - doesn't build with either available GCC or standard compiler, *NO* `dlopen()`

* Solaris 2.x/SunOS - nothing resembling even gcc 2.95 is available, `dlopen()` is broken and unsuitable for use in the ircd

* HP-UX - no `shl_load()` support in unrealircd, no build system support for HP's compiler

I do not think Unreal is in use on any of these platforms or ever has been...

tmcarthur:
Thanks for the information - I appreciate it. I still have some other things I need to check prior to removing this, but the information does make it easier to make a call. 

Thanks again!

PR #16, opened 2015-04-21
Backport from unreal 3.2: Introduce IsLoggedIn() and etc.

dboyz:
This allows IsLoggedIn() to be used in my previous pull request and fixes account name related issues as per 3.2. 
Cherry-pick'ed from 61f94bd6d87a499484e9854c7ac9a790a6e01ad4


syzop:
Implemented now (redid the patch though)

PR #17, opened 2015-04-21
Add missing permission error message as reported in rabbitircd

dboyz:
This is the exact commit I made to rabbitircd two years ago. Since the permission error is still missing, why not port it back to unrealircd. :)


tmcarthur:
Thanks for contribution!

PR #20, opened 2015-04-29
Something useful.

janicez:
Uh. Well, I noticed a compilation error in the Git alpha, decided to mess with permanent.c, noticed that delayjoin.c didn't really work properly, modified delayjoin.c to send hosts with unhiding users, and here you have, a pull request from your worst nightmare (formerly known as j4jackj).

Is there anything about this that is suspicious? This should work fine by my calculations.

### DIFF
--- a/src/modules/chanmodes/delayjoin.c
+++ b/src/modules/chanmodes/delayjoin.c
@@ -228,7 +228,7 @@ DLLFUNC void clear_user_invisible_announce(aChannel *chptr, aClient *sptr)
 	{
 		if (!is_skochanop(i->cptr,chptr) && i->cptr != sptr)
 			sendto_one(i->cptr,
-			    ":%s JOIN :%s", sptr->name, chptr->chname);
+			    ":%s!%s@%s JOIN :%s", sptr->name, sptr->user->username, GetHost(sptr), chptr->chname);
 	}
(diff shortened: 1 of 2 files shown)


tmcarthur:
Will have a look. Thanks!

tmcarthur:
Looks fine - will do a few quick tests and merge in - thanks.

janicez:
Well, um... GLINEs aren't at all working, and gline is a server only command. I'm not impressed. Just thought I'd point that out.

syzop:
GLINES not working? If that's so, feel free to report it on http://bugs.unrealircd.org/ and report the exact behavior so we keep track of these bugs. Issues mentioned in comments of random other issues are easy to loose track of.

You say you are not impressed? But by what? In general? Whether you know it or not, this is the first alpha release, so issues like these are entirely normal. That's why it's called alpha.. ;)

Just in case you unintentionally checked this branch out and was expecting a fully stable ircd, then use the branch 'master' instead, which contains UnrealIRCd 3.2.x.

Otherwise, if you want to help out and contribute to 3.4.x then feel free to do so! (as you already did)

tmcarthur:
G:line issue is fixed - let me know if you still encounter difficulties - thanks!

PR #22, opened 2015-05-19
Fix oper flag checks for global opers.

dboyz:
I am able to obtain the same umodes and oflags in both old and new codes which is very weird. **`OFLAG_ISGLOBAL`** in `m_oper.c` should be **`OFLAG_GLOBAL`** anyway. This PR may be wrong though.


tmcarthur:
Will have a look thanks!

dboyz:
Thanks @tmcarthur :)

syzop:
Oper flags are gone, so this can be closed.

PR #23, opened 2015-05-19
Add strip ANSI color codes ability in StripColors function. (#4101)

dboyz:
Code is not extensively tested. Variables nc, col, rgb and ans should be
an integer here.

### DIFF
--- a/src/modules/m_message.c
+++ b/src/modules/m_message.c
@@ -795,7 +795,8 @@ int size_string;
 
 char *_StripColors(unsigned char *text) {
 	int i = 0, len = strlen(text), save_len=0;
-	char nc = 0, col = 0, rgb = 0, *save_text=NULL;
+	int nc = 0, col = 0, rgb = 0, ans = 0;
+	char *save_text=NULL;
 	static unsigned char new_str[4096];
 
 	while (len > 0) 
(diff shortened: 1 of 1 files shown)


dboyz:
I referred to @grawity's irc-docs for ansi color code formats https://github.com/grawity/irc-docs/blob/b98b4bb4a36cba18a5440657d69027580af56ba5/client/epic5-colors2.txt

grawity:
I'm trying to understand what |nc| does here. Does it limit the number of `;` components or something?

(Actual color codes might have any number, from `\e[m` to `\e[;;;;m` to `\e[1;3;4;7;37m` to `\e[38;2;165;127;119m` to perhaps not-very-valid `\e[1;3;4;7;38;2;165;127;119;48;2;65;27;119m`...)

dboyz:
@grawity `nc` counts the number of characters after a color code is found. If special characters are found like `,` in normal color, `;` and `m`  in ansi color, the count resets to 0.

This code should work fine on your examples since `nc` is resetted each time when special character `;` is found and only stops *stripping* after special character `m` is found.

tmcarthur:
Will pull into my local branch and have a look - thanks.

dboyz:
Thanks! :)

syzop:
bump ;)

dboyz:
Revisiting my PR. Should we strip bold, italics and underline codes too in this function?

PR #25, opened 2015-05-20
Send UID/SID if available

dboyz:
There are dozens of other `sendto_server` codes lacking this u32 <-> u34 compatibility code.


syzop:
Patch is using CHECKPROTO() on the wrong client at all places.

It's not REQUIRED to send the NICKID (nothing will break if you don't), but yeah it's better. Perhaps some other time.

PR #27, opened 2015-05-26
Send RPL_HOSTHIDDEN on host changes

Zoddo:
This PR send a numeric 396 (RPL_HOSTHIDDEN) on sethost/chghost/vhost/+x/-x (as InspIRCd for example).

https://bugs.unrealircd.org/view.php?id=4224


tmcarthur:
Thanks - will have a look!

tmcarthur:
Looks good - merging in.

PR #30, opened 2015-07-14
Use "NOTICE *" before registration instead of "NOTICE AUTH".

sadiepowell:
AUTH is a valid nickname so sending notices to it is probably not a good idea. Use \* as the target instead as done with numerics when the nick is not available.

This mimics the behaviour in Charybdis, IRCD-Hybrid, InspIRCd 2.2, Plexus 4, etc.


tmcarthur:
Thanks for contribution!

PR #34, opened 2015-09-11
sendto_serv_butone_nickcmd: fix nickv2 format

Adam-:


syzop:
Thanks for the report Adam. nenolod ripped out several capability checks, sending some things "always" and some things "never". I did some de-duplication of the code and fixed this.

PR #38, opened 2015-10-17
Fork safety

kaniini:
This pull request moves fork() to a logically better place and removes several other problems associated with the current location of fork(), by ensuring that it is at a point in the program where no asynchronous failure conditions can occur.

Additionally, it has been load tested on Windows, Mac OS X, Linux (both glibc and musl), FreeBSD and Solaris.

It is an extremely low risk merge.

Signed-off-by: William Pitcock <nenolod@dereferenced.org>


tmcarthur:
We can consider for 4.1 - not for 4.0

PR #39, opened 2015-11-02
fix: you should put pid into $PID_FILE

jackywu:
you should put pid into $PID_FILE, otherwise, any other command related with $PID_FILE will fail.


syzop:
Thanks for the pull request but the pid is written by UnrealIRCd (bin/unrealircd) itself and not by the startup script. So this is not a correct change.
In case you are encountering some kind of bug, feel free to report it at https://bugs.unrealircd.org/ (FYI pid writing/reading is working fine for me)

PR #42, opened 2015-12-05
m_nouserctcp: module which blocks all CTCP from users

esialb:
m_nouserctcp: module which blocks all CTCP from users, not just CTCP sent to channels, and disconnects any user attempting to use CTCP.

Useful for preventing accidental information leakage between clients, such as by CTCP VERSION or CTCP PING messages.


esialb:
i guess the version on the plugin should be changed to something other than that of m_noctcp

syzop:
I think it would be better to change the current user mode +T to also block channel CTCP's (this would require some core changes though). That would be more logical.

esialb:
Would user mode +T prevent receiving or sending CTCP?

syzop:
I've added the feature request (the one I described) at https://bugs.unrealircd.org/view.php?id=4492

I'm closing this pull request as the feature on bugs* would come down to the same thing, except receiver/sender are swapped but that's only for the better IMO, and is configurable via a user mode.

PR #43, opened 2015-12-10
Allow shunned users to use the PING command

fduchene:
When a user is shunned (eg /tempshun user ), the command PING cannot be used (while PONG can, to answer server PING).
Some clients like irssi are using PING  command to compute the server lag, so when an irssi user is shunned, the lag displayed in irssi start to increase, giving the user a way to know if he is shunned.
After 320 sec of lag, irssi will reconnect, bypassing automatically the tempshun.


syzop:
Will seriously consider this after 4.0.0 freeze

Zoddo:
:+1:

PR #44, opened 2015-12-12
build: Install the run script in the base dir

staticfox:
As the readme suggests, by default place the start/stop
script in the same directory as the install

---
Per README:
`Run ./unrealircd start in the directory where you installed UnrealIRCd.`
However, it is installed in $HOME unless --with-scriptdir is passed to the configure script.
Also, under Makefile.in, I changed it so copy/pasting what `To start/stop UnrealIRCd run:` spits out will run. Apologies for the whitespace cleanups, my editor automatically does that :x. I can undo it if you'd like.

### DIFF
--- a/Config
+++ b/Config
@@ -3,7 +3,7 @@
 # Config script for UnrealIRCd
 # (C) 2001 The UnrealIRCd Team
 #
-# This configure script is free software; the UnrealIRCd Team gives 
+# This configure script is free software; the UnrealIRCd Team gives
 # unlimited permission to copy, distribute and modify as long as the
 # copyright headers stay intact
 #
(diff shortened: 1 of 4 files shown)


syzop:
Thanks for the report.
As for the pull request, yes if you could remove the whitespace changes next time you submit something it's easier to read ;).
I've narrowed it down to a 1 line change:
https://github.com/unrealircd/unrealircd/commit/49384b89039457e1429684e17972f21719fc4210
[unreal40 49384b8] The 'unrealircd' script was missing if you installed to a directory other than ~/unrealircd. Reported by staticfox.

But again, thanks for the report, silly bug.. as you may have figured out I always install to ~/unrealircd so I never noticed myself :)

PR #45, opened 2015-12-28
Add mirc scripts to allow block.

sketchni:
Also fixed a typo that was bothering me. Sorry ;)


syzop:
I don't think we should whitelist (=consider safe) .ini files and .mrc files by default, it's not like mIRC scripts cannot have virusses.
The ~O extban was missing earlier indeed and has been added some time ago (though at the proper place ;p).

PR #46, opened 2016-01-03
Add ability to hide quit messages of K/G/Z:lined users

Zoddo:
This add `set::options::secret-ban-reasons` in the configuration file

Suggested by Aubrey ([#3993](https://bugs.unrealircd.org/view.php?id=3993))

### DIFF
--- a/doc/conf/examples/example.conf
+++ b/doc/conf/examples/example.conf
@@ -390,6 +390,7 @@ set {
 	options {
 		hide-ulines; /* hide U-lines in /MAP and /LINKS */
 		show-connect-info; /* show "looking up your hostname" messages on connect */
+		/* secret-ban-reasons; don't publicly display K/G/Z:lines reasons (in the QUIT message) */
 	};
 
 	maxchannelsperuser 10; /* maximum number of channels a user may /JOIN */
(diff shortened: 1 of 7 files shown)


doc/conf/examples/example.conf
```diff
@@ -390,6 +390,7 @@ set {
 	options {
 		hide-ulines; /* hide U-lines in /MAP and /LINKS */
 		show-connect-info; /* show "looking up your hostname" messages on connect */
+		/* secret-ban-reasons; don't publicely display K/G/Z:lines reasons (in the QUIT message) */
```
GLolol: Publicly, not public**e**ly.

Zoddo: Oops, corrected :)

dboyz:
I might be wrong but I think you may want to initialize `secret_ban_reasons` somewhere.

Zoddo:
It seem there isn't other code for the other config options (I [checked](https://github.com/unrealircd/unrealircd/search?utf8=%E2%9C%93&q=allow_insane_bans) `allow_insane_bans`)

syzop:
See https://bugs.unrealircd.org/view.php?id=3993 for discussion

PR #47, opened 2016-01-13
use #include "unrealircd.h" in more files

blank4:


syzop:
Thanks for the patch. Would have been better if you test-compiled, though. This patch broke the build process. (Fixed now)

cehib:
@syzop what did you fix as the merged files are exactly the same as what I originally done?

syzop:
Your patch (and indeed, the commit) caused compile errors in both censor modules.
I fixed this 1 commit later, in https://github.com/unrealircd/unrealircd/commit/d540d652518122c8ebd35c084ec254afa2c43c24

PR #49, opened 2016-03-05
Add an option to hide the notices about TLS parameters

grawity:
I don't like my servers being overly verbose. The fingerprint can be seen in one's own /whois anyway.

### DIFF
--- a/include/dynconf.h
+++ b/include/dynconf.h
@@ -61,6 +61,7 @@ struct zConfiguration {
 	unsigned ident_check:1;
 	unsigned fail_oper_warn:1;
 	unsigned show_connect_info:1;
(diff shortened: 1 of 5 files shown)


sketchni:
I'd like to weigh in that this would be a nice thing to be able to disable. Enabled by default would be fine.

syzop:
Thanks for the patch! I think it's a useful setting.
Don't totally like the name of the new setting but I don't want to spend more time on this either ;)

grawity:
I'd have used `show-connect-ssl-info` myself, but I decided to avoid changing defaults for now.

(Maybe it could have been `show-connect-ssl-info yes/no;`. Oh well.)

PR #50, opened 2016-03-08
Reduce required settings

grawity:


syzop:
I agree. Thanks for the PR.

PR #52, opened 2016-03-19
Fixed an issue when TOPICLEN should be AWAYLEN

westor7:
Renamed the TOPICLEN into AWAYLEN to have the correct value.


syzop:
Hmm.. does this even compile? I don't see AWAYLEN being defined in src/ or include/

syzop:
Not important and PR creates a build failure. Closing...

PR #53, opened 2016-04-24
Update m_join.c

SketchNI:
Line #454 sends as a notice rather than text.

### DIFF
--- a/src/modules/m_join.c
+++ b/src/modules/m_join.c
@@ -448,7 +448,11 @@ CMD_FUNC(_do_join)
 								get_client_name(sptr, 1), name);
 						}
 						if (d->reason)
-							sendto_one(sptr, err_str(ERR_FORBIDDENCHANNEL), me.name, sptr->name, name, d->reason);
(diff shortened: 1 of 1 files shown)


GLolol:
The numerics for "No such channel" and others are defined by RFC 1459: https://www.alien.net.au/irc/irc2numerics.html. Removing them is going to subtly break bots and clients depending on standard behaviour.

sketchni:
A little bit. :stuck_out_tongue: 

Basically what I've done is the difference between:  
````#chan Channel is forbidden: No channels can be created.````
and
````[11:04:10pm] -example.net- *** Can not join #chan: No channels can be created.````

syzop:
The general idea is to use more numerics and less notices, this so clients can interpret the text and show things in a better way. This change would do the opposite (even though it's meant as in the users best interest). So I'm afraid this change won't be accepted.

I did another change though, I changed the "is forbidden" to "cannot join channel":
/* 448  ERR_FORBIDDENCHANNEL */ ":%s 448 %s %s :Cannot join channel: %s",

Which seems more appropriate. Especially for the case you suggested (denying all channels except a few).

PR #54, opened 2016-04-27
Update curl-ca-bundle.crt

sketchni:
Add in Lets Encrypt Authority. This is needed for remote includes (at least) when a webserver uses Lets Encrypt certificates.


syzop:
Websites using Let's Encrypt certificates work fine with current ca bundle, but the site needs to configure the certificates correctly. That means: with the intermediate certificates (also called certificate chain or bundle). This is because the Let's Encrypt certificate is cross-signed with an IdentTrust certificate, and this latter is present in the CA bundle we ship (and many browsers, for that matter). See https://letsencrypt.org/certificates/ for more technical information if it interests you.

You can check yourself via something like:
`wget --ca-certificate=curl-ca-bundle.crt -O /dev/null https://letsencrypt.org/certificates/`
Which succeeds with UnrealIRCd 4.0.3's CA bundle.

Apparently you are having problems with some site that uses Let's Encrypt? You can use https://www.sslshopper.com/ssl-checker.html to check the site to debug the problem. You will probably see it's missing an intermediate certificate. (How to fix that is beyond this pull request but you should find plenty of information on the internet about it)

sketchni:
Yep, it was a missing intermediate cert. I'll close the PR.

PR #41, opened 2015-11-26
Correcting a tenacious typo. s/succesfully/successfully/g

theRoda:
I am writing grok filters for unreal logs and having to intentionally type this typo is figuratively killing me.

PR #55, opened 2016-05-09
Fix typos ("success" instead of "succes")

GLolol:
Original patch by @theRoda. Closes #41.


syzop:
I've applied your patch, except for the changes to Changes.old / Changes.older since I don't think we should edit these anymore.
Done (https://github.com/unrealircd/unrealircd/commit/64839a6304230d01856f347a2589b58f7d711956). Thanks for the patch!

PR #58, opened 2016-08-17
Use localstatedir instead of datadir for tunefile, etc.

binki:
datadir is normally /usr/share and specifies the location of files
that are never modified. localstatedir is normally /var/lib and
where you are supposed to place files which should be modified.
This change moves slightly towards doing directories how one
should when using autoconf and uses the existing --localstatedir
option, removing the redundant and misused --with-datadir.

The change should be transparent to anyone using ./Config. I have
also added --with-standard-dirs which would give one directly
invoking ./configure and passing --localstatedir the results
they expect (if you passed --localstatedir=/var/lib you’d expect
unrealircd to store its state in /var/lib/unrealircd).

### DIFF
--- a/Config
+++ b/Config
@@ -72,7 +72,7 @@ ARG="$ARG--disable-prefixaq "
 fi
 
 ARG="$ARG--with-bindir=$BINDIR "
-ARG="$ARG--with-datadir=$DATADIR "
+ARG="$ARG--localstatedir=$DATADIR "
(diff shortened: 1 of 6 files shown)


syzop:
Closing some old PR's.
It was an intentional decision not to use the standard dirs, for more information see the bugnote from June 2015, in particular under "I tried the defaults"..: https://bugs.unrealircd.org/view.php?id=4361#c18407

PR #59, opened 2016-08-22
Added server count to the local server notice for /rehash -global

GottemHams:
In regards to my [bug report](https://bugs.unrealircd.org/view.php?id=4739) I figured I may as well give it a shot myself.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+._.DS_Store
+.DS_Store
+
 # Ignore configure step
 config.log
 conftest.*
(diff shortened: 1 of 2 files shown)


bonnedav:
Why has this not been merged yet?

syzop:
If this would be changed to be a single patch that only does what https://bugs.unrealircd.org/view.php?id=4739 suggested then it can be merged (if it's good). Right now it consists of 3 patches with more unrelated than related changes :D
Actually I'm just going to close it, best to resubmit I think.

GottemHams:
Yeahhhhhh I think I made this pull request before I even wrote my first module, so. :D I'll redo it sometime. Also not to mention I set up my Sublime Text to strip trailing whitespace.

GottemHams:
Better yet, I'm going to let go of this "patch" entirely. :D I had to add a custom rehash -flag due to a certain module I'm using on our network and I simply added the functionality in there. =]

PR #60, opened 2016-08-25
Generation and relative path of DH parameters

capitaine-s:
I've reported it https://bugs.unrealircd.org/view.php?id=4738


dboyz:
Seems to be correct but I have not tested the code.

Thanks for the PR! :+1:

syzop:
Thanks, fixed! 603adb8ec21643f7a129b29b782ca4ba0e676fdc

PR #61, opened 2016-09-11
Fix ./Config SSL prmopt when previously nonempty for bug #4749.

binki:
https://bugs.unrealircd.org/view.php?id=4749


syzop:
Thanks, fixed a7f98af6ed9f17436fae68774a8a8b2070a58bb7

PR #63, opened 2016-12-07
hunt_server: allow hunting servers by wildcard

Adam-:


bonnedav:
Why has this not been merged yet?

syzop:
Closing some old pull requests...
This was an intentional behavior change, see https://bugs.unrealircd.org/view.php?id=4482

jlu5:
I respectfully disagree - I think this is a significant regression from 3.2, and many other IRCds I've messed with in the past support the same feature..

PR #66, opened 2016-12-22
Remove 'gay' from the bad words list.

sadiepowell:
Cause this was an embarrassingly archaic view a decade ago never mind in 2016.


syzop:
The term "gay" is often used in a derogatory way ("gay this", "gay that"). I'm not sure much has changed in that way since 2003. I wish I could agree with you but I have my doubts, still seeing the term used that way from time to time. In any case, it sounds odd to censor the word when it may appear in a perfectly normal sentence. UnrealIRCd cannot guess the "context" of a conversation after all.
I'm going to remove the word from the list.

PR #68, opened 2017-07-06
Support --without-privatelibdir for packagers.

binki:


binki:
I’m leaving running `./autogen.sh` up to you, @syzop.

PR #71, opened 2018-04-29
Enable tests for OS X builds

ghost:


syzop:
Closing some old PR's.
I'm just going to close this. The alternative was to keep it lingering around until it actually works, but it has been 6 months now, so.. ;)

PR #72, opened 2018-05-13
Update operclass.default.conf

TehPeGaSuS:
IMHO, I think that `restart` and `die` privileges should be enabled by default to network administrators. This fix aims to correct that.
If, however, you feel that things should be kept as is, feel free to ignore this PR.

Cheers,
The_Myth

### DIFF
--- a/doc/conf/operclass.default.conf
+++ b/doc/conf/operclass.default.conf
@@ -115,7 +115,7 @@ operclass netadmin {
 		immune;
 		notice;
 		self;
-		server { opermotd; info; close; remote; module; dns; addline; rehash; description; addmotd; addomotd; tsctl; };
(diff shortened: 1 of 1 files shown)


acidvegas:
Well on SuperNETs, we have a hub and a collection of leaf servers that connect to the hub. All the leaf servers use remote includes for their config. Any changes made can mostly be applied with a REHASH. If you are make a change to a specific server, chances are you are already on the box its on and can `./unrealrcd restart`

I have just never used DIE or RESTART ever in my time of being a network admin.

bonnedav:
I think netadmins should have all server permissions, by changing the line to just `server;`.

syzop:
Closing some old PR's...
This was discussed and rejected in https://bugs.unrealircd.org/view.php?id=4512
For more information, see that bug report.

PR #74, opened 2018-09-22
Fix for GCC > 7.3 building

ktrace:
source of patch: 
https://github.com/calculatelinux/calculate/commit/1886e7a045667f192b01c893500959a445144e45


syzop:
"wrong branch, use #75 instead" (3.2.x is EOL)

PR #75, opened 2018-09-22
Fix for GCC > 7.3 building

ktrace:
source of patch:
https://github.com/calculatelinux/calculate/commit/1886e7a045667f192b01c893500959a445144e45

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -155,7 +155,7 @@ void ban_flooder(aClient *cptr)
 /*
  * This routine adds fake lag if needed.
  */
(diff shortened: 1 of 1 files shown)


syzop:
Hmm, what error are you seeing with GCC>7.3 on UnrealIRCd 4.x that needs to be fixed?

syzop:
I think this can be closed. The fix is unnecessary for 4.x. From what I read on IRC this was an attempt to get a fix backported to 3.2.x - which is end of life for about 2 years so that won't be done.

PR #79, opened 2019-02-01
 Add "/BLCHECK ip" command to manually check blacklists

k4bek4be:


syzop:
The idea itself is fine but I don't like the way it is implemented. Sorry.
I can have a go at this another time. I suggest you don't work at it in the meantime.

PR #80, opened 2019-02-10
Update example.conf

TehPeGaSuS:
Added `set::modes-on-join "+nt"` to the current configuration to help users to keep their channel with sane defaults, when joining a non registered channel
Added `set::restrict-usermodes "x"` as being able to disable the default cloaking is a security risk
Changed `set::allow-userhost-change` to `force-rejoin` since the default `always` may lead to clients desynch
Added `set::cloak-method ip` since that way whe can still have the XXX.YYY.ZZZ.IP cloak without the need to disable DNS resolving
Added `set::options::identd-check` since it barely interfere with user connection and is helpful for public services that are properly configured, such as ZNC

That's all for now xD

Regards,
PeGaSuS


syzop:
Moved to https://bugs.unrealircd.org/view.php?id=5274

PR #81, opened 2019-04-20
Helpop update

TehPeGaSuS:
Added soft actions help on `/helpop gline` and `/helpop kline`.
Changed `*:Line(s)` to be `*-Line(s)` which are more consistent with the new IRC messages


TehPeGaSuS:
And what about this one? :D

PR #84, opened 2019-11-03
Update RELEASE-NOTES.md

TehPeGaSuS:
Let's not forget `git` users, heh! xD


syzop:
Thanks, I appreciate the feedback.
I think, however, that if you run git then you are smart enough to understand the original phrase that you should run at least version X. Changing the sentence to include those few % users that run git (and which, as said, I think already understand) also has the disadvantage of making the sentence unnecessary long.

PR #85, opened 2019-11-06
Enhance Windows Security: Changes to unrealsvc.c and gui.c which make it possible to run Unreal…

AlexandraBryant:
**Makes it possible to run UnrealIRCd as a service on Windows with non-admin permissions, in line with the official documentation which recommends Linux users [NOT run as Root](https://www.unrealircd.org/docs/Do_not_run_as_root).** 

**unrealsvc.c**
•	Adjusted service access level from `SERVICE_ALL_ACCESS` to `SERVICE_CHANGE_CONFIG`
•	Adjusted the account to log on as to `“NT AUTHORITY\\NetworkService”`
**gui.c**
•	Adjusted the access level to the service control manager from `GENERIC_EXECUTE` to `SC_MANAGER_CONNECT`

Expands on the solution where another user [encountered this problem in the past](https://forums.unrealircd.org/viewtopic.php?t=3476). Now no workarounds or 'heinously complex' permissions would be needed, it's simply runs as Network Service.

### DIFF
--- a/src/win32/gui.c
+++ b/src/win32/gui.c
@@ -218,8 +218,8 @@ int APIENTRY WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLi
 	GetOSName(OSName);
 	if (VerInfo.dwPlatformId == VER_PLATFORM_WIN32_NT) 
 	{
-		SC_HANDLE hService, hSCManager = OpenSCManager(NULL, NULL, GENERIC_EXECUTE);
-		if ((hService = OpenService(hSCManager, "UnrealIRCd", GENERIC_EXECUTE))) 
+		SC_HANDLE hService, hSCManager = OpenSCManager(NULL, NULL, SC_MANAGER_CONNECT);
+		if ((hService = OpenService(hSCManager, "UnrealIRCd", SC_MANAGER_CONNECT))) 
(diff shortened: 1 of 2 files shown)


syzop:
Right, so this would be a patch for https://bugs.unrealircd.org/view.php?id=2330 ?
I will take a look at it in the next few days. If it works well then I'll put it in. Thanks!

syzop:
Thanks! Done in UnrealIRCd 5.
See https://bugs.unrealircd.org/view.php?id=2330 for the entire story :).

AlexandraBryant:
Hi @syzop, would it be possible to include the security patch in the 42 branch?

syzop:
With UnrealIRCd 5 just around the corner (this year, so within 7 weeks) I won't be adding this to 4.2.x. The final patch is a lot longer and has a bit more impact. I don't want to risk doing that in the current stable tree. Also, this is a security enhancement, not so much as a fix for an unintentional mistake (even apache runs as SYSTEM on Windows by default). It's our policy to only do bug fixes in the 4.2.x branch since May 2019. See https://www.unrealircd.org/docs/UnrealIRCd_releases

So yeah, got to wait for UnrealIRCd 5, but that will be relatively soon! :)

Hope you understand.

PR #87, opened 2020-01-01
Two quick fixes with the module manager.

k4bek4be:


syzop:
I'm a bit unsure whether to commit this. You updated it so the errors/warnings match the line numbers in the source file. That is logical if you think one way.
Right now it prints out line numbers of within that configuration block, and that block is printed out under it, so when it says line 3 there it is line 3 there. That is logical in another way.

k4bek4be:
My point is that the editors i use to edit sources usually display line numbers in a file, and the command dumping the config block does not. And i, encountering a missing ; character, hadn't thought about counting the lines manually there, instead repeatedly modifying a (good) line in the module source to see what's going on, only to figure out the problem is way below.

syzop:
We could do it your way, but then the dump of the config block after "configuration block within .." should be removed, since otherwise it would only be confusing in my eyes.

PR #89, opened 2020-01-04
Add support for "webirc-bypass yes;" in connthrottle module

westor7:
This is my second attempt to add "webirc-bypass yes;" in config in order to give exceptions in all WEBIRC users, it seems useful in certain cases.

I left it enabled by default if the config option doesn't exist, it seems normally to allow WEBIRC users to connect while the module is working, that's only my personal opinion about this.

This closes one of the 3 suggestions in https://bugs.unrealircd.org/view.php?id=5506

### DIFF
--- a/src/modules/connthrottle.c
+++ b/src/modules/connthrottle.c
@@ -1,6 +1,6 @@
 /*
  * connthrottle - Connection throttler
- * (C) Copyright 2004-2019 Bram Matthys (Syzop) and the UnrealIRCd team
+ * (C) Copyright 2004-2020 Bram Matthys (Syzop) and the UnrealIRCd team
  * License: GPLv2
  * See https://www.unrealircd.org/docs/Connthrottle
  */
@@ -91,7 +91,7 @@ MOD_TEST()
 	safe_strdup(cfg.reason, "Throttled: Too many users trying to connect, please wait a while and try again");
 	cfg.minimum_reputation_score = 24;
 	cfg.sasl_bypass = 1;
-	cfg.webirc_bypass = 1; /* I am not sure... probably disabled with 0 by default? */
+	cfg.webirc_bypass = 0;
 
 	HookAdd(modinfo->handle, HOOKTYPE_CONFIGTEST, 0, ct_config_test);
 	HookAdd(modinfo->handle, HOOKTYPE_CONFIGPOSTTEST, 0, ct_config_posttest);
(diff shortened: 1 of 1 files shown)


src/modules/connthrottle.c
```diff
@@ -355,10 +363,11 @@ EVENT(connthrottle_evt)
 	if (ucounter->rejected_clients)
 	{
 		snprintf(buf, sizeof(buf),
-		         "[ConnThrottle] Stats for this server past 60 secs: Connections rejected: %d. Accepted: %d known user(s), %d SASL and %d new user(s).",
+		         "[ConnThrottle] Stats for this server past 60 secs: Connections rejected: %d. Accepted: %d known user(s), %d SASL %d WEBIRC and %d new user(s).",
```
thesamesam: Add a comma here?

src/modules/connthrottle.c
```diff
@@ -89,6 +91,7 @@ MOD_TEST()
 	safe_strdup(cfg.reason, "Throttled: Too many users trying to connect, please wait a while and try again");
 	cfg.minimum_reputation_score = 24;
 	cfg.sasl_bypass = 1;
+	cfg.webirc_bypass = 1; /* I am not sure... probably disabled with 0 by default? */
```
thesamesam: AFAIK this doesn't actually set the default, it just tests the config if the default was 1? Not sure.

westor7: I think actually that MOD_TEST() is just looking in the CFG and if there is missing any of the available option it uses a default one, i left it as 1 value to enable it in that case, i don't know if the unrealircd team agree with that or not.

src/modules/connthrottle.c
westor7: I hate commas :P

src/modules/connthrottle.c
syzop: Yeah, these set the defaults. I usually put them in a config_defaults() function to make it more clear. In this case it has a clear comment above it though: ```/* Defaults */```. Perhaps you missed it because of the short context in the diff from GitHub.

syzop:
Hi westor, could you resolve that 'comma' issue and also I would like the webirc bypass to be NOT the default. With those 2 changes I think it can be merged.
You should be able to simply make updates to your git branch and git commit & git push and the changes should appear here in the PR automatically.

PR #91, opened 2020-01-15
Fix an incorrect /ELINE example bantype

westor7:
It should be 'G' instead of 'g' in the /ELINE example


westor7:
I will open a new including one more fix.

PR #93, opened 2020-01-29
Add [IP] tag on nick change in server notice

westor7:
This is going to display the nickname IP address in server notice (same as when a user connecting/exiting) it is useful for some cases.

Before:
*** old_nick (ident@host) has changed their nickname to new_nick

Now:
*** old_nick (ident@host) [127.0.01] has changed their nickname to new_nick


syzop:
The patch itself is fine, but I would like to make it a larger change, see https://bugs.unrealircd.org/view.php?id=5548 at some later stage.
So I'm closing this PR and we'll revisit / look at those other changes later this year.

PR #95, opened 2020-02-14
Allow modules to implement fully-featured TAGMSG without resorting to…

k4bek4be:
… strange hacks


syzop:
We will implement ```TAGMSG``` in UnrealIRCd somewhere in the next few months. I don't think it would be correct to pull in such a change until then. For those who cannot wait a couple of months and who want to use your module, well.. sounds like you already worked around the problem.

Shillos:
Well, that is not bad to have a workaround even temporarily. It works ok on my server. Hope the wait will be soon because there are lots to be done. I saw you have some `TAGMSG` in the code, so this shows some hope and promissing for the near future.

PR #96, opened 2020-02-17
Add invite-notify cap

k4bek4be:


syzop:
This does not work correctly for remote clients (the same mistake neno made years ago).

I think this is likely the least most interesting IRCv3 spec though, so I don't have much interest coding it myself at the moment.

syzop:
This doesn't work without major changes. k4be will create a separate patch first for the API change. Then invite-notify can always be done later.

PR #98, opened 2020-03-09
Removing duplicate line in conf prog header

Ubunfu:
Noticed this duplicate line printed in the configuration program header while building and installing from source for the first time today.  It appears when you run the ```./Config``` [build step](https://www.unrealircd.org/docs/Installing_from_source#Compiling).  I know it's very small, but thought I'd submit a PR anyway since it's quite user-facing.


syzop:
Funny, never noticed it. Thanks!

PR #100, opened 2020-03-23
Fix trailing space in RPL_NAMREPLY

delthas:
Currently, UnrealIRCd outputs a trailing space to the trailing parameter of RPL_NAMREPLY (when there are users in the channel). However, according to RFC 1459 and RFC 2812, this is forbidden, since the grammar for this reply is (quoting RFC 2812):
```
( "=" / "*" / "@" ) <channel> :[ "@" / "+" ] <nick> *( " " [ "@" / "+" ] <nick> )
```

This commit fixes the issue by adding the joining space only if there is another, remaining nick to be appended to the buffer.

### DIFF
--- a/src/modules/names.c
+++ b/src/modules/names.c
@@ -183,7 +183,8 @@ CMD_FUNC(cmd_names)
 		 */
 		for (; *s; s++)
 			buf[idx++] = *s;
-		buf[idx++] = ' ';
+		if (cm->next)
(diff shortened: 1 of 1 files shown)


syzop:
I think if your parser fails on these kind of issues, you will have a very hard time dealing with IRC servers.... of various brands, not just us.
The patch looks OK, though, and is not intrusive, so I will merge it.

PR #110, opened 2020-05-25
Add more places to auto detect for OpenSSL

westor7:
Add /usr/bin/openssl to auto detecting for OpenSSL on the machine too.

### DIFF
--- a/configure
+++ b/configure
@@ -1402,7 +1402,7 @@ Optional Features:
   --enable-hardening      Enable compiler and linker options to frustrate
                           memory corruption exploits [yes]
   --disable-prefixaq      Disable chanadmin (+a) and chanowner (+q) prefixes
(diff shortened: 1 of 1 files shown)


syzop:
It already detects OpenSSL in ```/usr```. Which isn't that surprising, since that is where 99% of the people have it anyway :).

The way that list works is that it is a directory, then it checks ```<dir>/bin/openssl``` and ```<dir>/include/openssl``` and such. As you can see the entry for ```/usr``` is already there at the end.

What you did is make it check for ```/usr/bin/openssl/bin/openssl``` and ```/usr/bin/include/openssl``` and that sort of thing. That is not what we want ;)

syzop:
In any case, there must be a reason why you submitted this. So you probably had some sort of problem where it failed to detect OpenSSL? We can also discuss on IRC.

westor7:
You're right my PR is incorrect, apologize.

PR #112, opened 2020-05-28
Call user_account_login on UID message too

k4bek4be:


syzop:
Where is this for? The title of this PR is basically what the code does, but is missing the "why" or purpose :D
(And the description is empty)

k4bek4be:
One of my modules (metadata-db) is designed to act on each case of user login. One, and probably most common, of these is a successful pre-registration SASL authentication. That is handled by the user's local server, and later broadcast by the UID command to other servers.
Other cases, already handled neatly, are the SVSMODE message (when user has logged in after registration, using SASL or nickserv message) and SVSLOGIN message (local unregistered connection that used SASL successfully).

syzop:
Ah, right, so also call the hook for remote logins, for modules.

syzop:
Hmm.. I have some doubts about this. This would also mean that you would get like hundreds of HOOKTYPE_ACCOUNT_LOGIN events just when two servers, sync, right?
I am not sure if it is "correct" in such a case, so when server syncing, to call HOOKTYPE_ACCOUNT_LOGIN, since the user actually isn't logging in to his/her account, it are just servers syncing.
Shouldn't we skip if !IsSynched(client->srvptr) ?

syzop:
.. then again, we also call things like HOOKTYPE_AWAY in such a case, so maybe I shouldn't care.
A module can decide itself if it wants to skip such events or not, based on IsSynched(client->srvptr)

PR #113, opened 2020-05-30
Fix to display the TLSversion correct on connect

westor7:
Previously it didn't display correctly on server notice the TLSv* version on local connection.

Example:

**TLS_CHACHA20_POLY1305_SHA256**

Should be displayed as: **TLSv1.3-TLS_CHACHA20_POLY1305_SHA256**


westor7:
Credits to Jobe :D

PR #114, opened 2020-05-30
Add ciphers value on RPL_WHOISSECURE (671) locally

westor7:
If someone want's to view the TLS version + ciphers from a local client SSL connection he would now by WHOIS the user, this only will work for local clients and not for remote clients.

If this is security privacy risk addition then do not proceed of course.

Credits to Jobe who created the code.

- Thanks


westor7:
BAD PR , APOLOGIZING!!!

PR #115, opened 2020-05-30
Add ciphers value on RPL_WHOISSECURE (671) locally

westor7:
If someone want's to view the TLS version + ciphers from a local client SSL connection he would now by WHOIS the user, this only will work for local clients and not for remote clients.

If this is security privacy risk addition then do not proceed of course.

Credits to **Jobe** who created the code.

- Thanks!

### DIFF
--- a/src/modules/whois.c
+++ b/src/modules/whois.c
@@ -314,7 +314,7 @@ CMD_FUNC(cmd_whois)
                                 if (MyUser(target))
                                         sendnumericfmt(client, RPL_WHOISSECURE,
                                                 "%s :is using a Secure Connection (%s)", target->name,
-                                                tls_get_cipher(target->local-ssl));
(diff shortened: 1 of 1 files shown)


TehPeGaSuS:
I like the idea, but this should also be applied to remote connections.
It is useful to see what TLS ciphers are commonly used and set stricter ciphers rules on `set::tls::ciphers`.

syzop:
UnrealIRCd now has this (even for remote connections). IRCOps can see the ciphers. Users cannot by default but you can change the [set::whois-details](https://www.unrealircd.org/docs/Set_block#set::whois-details) to show it to self or to all other users too, whatever you wish!

I'm not considering this for U5, so closing this one. Thanks for the suggestion though, without it this would not have been in U6 either.

PR #116, opened 2020-06-09
Fix problem with labeled-response on websocket connections

k4bek4be:


syzop:
Thanks, fixed now in https://github.com/unrealircd/unrealircd/commit/dcb89f933e10af3148bde9188308ba853ab907ff

PR #117, opened 2020-06-18
Update help.conf

TehPeGaSuS:
In fact, the umode +T prevents other users from sending us CTCPs and not the other way around ;)


syzop:
I'll take the suggestion with me during a partial rewrite, since there is a bigger picture here. https://bugs.unrealircd.org/view.php?id=5814.
Thanks for bringing it up.

PR #118, opened 2020-07-26
Remove 435 BANNICKCHANGE support

ddevault:
This was only ever of dubious legitimacy and it interacts poorly with
qbans (quiets). For example, a channel which only wants registered users
to speak may set +q ~q:~a:*. If there are a problem with services, or
someone reconnects and auto-joins many channels, they will have to
manually part and rejoin any channels using this approach before they
can change their nick.

This also removes the ~n extban, under a rationale which applies to
both: global settings like the user's nick are beyond the channel
operator's domain of privilege to control of their channel members.

Even if one holds these features to be legitimate, they are trivially
subverted - leave the channel, change your nick, then return - and
therefore are not justified in the face of the problems they introduce.
If someone is using nick changes to subvert a mute of another sort, then
an affected operator should naturally escalate the mute to a ban to
correct the problem.

The global channel mode +N has been left in, as it's only ever used as a
temporary measure against floods and is not considered reasonable to
leave on permanently.


syzop:
>  For example, a channel which only wants registered users to speak may set +q ~q:~a:*

In UnrealIRCd if you want only registered users to speak you would set ```+M```, a channel mode added 17 years ago which works fine for this purpose. It won't have any of the side effects you mention with regards to nick changing.
In fact ```+b ~q:~a:*``` will never work as we don't process wildcards in ~a. It sounds like you wrote some kind of generic comment that applies to other IRCds, as the problem you mention (use of ~q:~a with certain side effects) does not exist in UnrealIRCd.

I understand what you are saying with regards to scope and such of channel settings impacting things like inability to change a nick name. But, as you also said, sometimes this is needed like when setting +N in case of a flood (may even be done automatically via +f). We do not see abuse or annoying effects of this in real life, as said the ~q:~a: case does not apply to us, we have +M for that. And from what I can gather, whenever a ~n is set (which is rare), it is set specifically for some troublemaker (or more often: troublesome IRC client or bouncer) with the intention of only blocking that nick change and thus a "lesser" effect, otherwise that user would have been banned.

Disabling target bans by default is not what I would suggest. It also does not work like you say "simply part, change and rejoin", it would be "simply part, change and then find out that you are banned". Preventing that is the whole point of target ban.

Even with all of the above in mind, you can still configure UnrealIRCd entirely according to your wishes, we allow a lot of tweaking:
1) Use ```set { check-target-nick-bans no; }```
2) Don't load (or blacklist) the ```extbans/nickchange``` module
3) Optionally, not recommended but if you want this, don't load (or blacklist) the ```chanmodes/nonickchange``` module

ddevault:
You're correct that I'm coming here from another ircd. The problem will affect unrealircd servers running atheme, though, by preventing nickserv regain from working. Consider freenode: despite `+r` being available for 17 years (or whatever it is), people still quiet `$~a`, leading to this problem. unrealircd users can likely be expected to set `+b ~q:~a` as well.

Anyway, since you accurately surmised that I came here trying to proliferate a change I'm arguing for in another ircd, I don't have a horse in the unrealircd race and I'm not especially interested in arguing the feature.

KoraggKnightWolf:
Except that UnrealIRCd users CANNOT be expected to set +b ~q:~a as was said since this DOES NOT EXIST in UnrealIRCd. Do not just blatantly assume every IRCd has the exact same extbans/modes/commands etc and jump across projects wondering why your idea gets rejected. Ripping out the ability to block nick changes just opens up spammers to do NICK change floods. Your alternative approach to Atheme which bypasses the race condition is a more suitable way of handling the situation.

ddevault:
Except that it does:

https://www.unrealircd.org/docs/Extended_bans

Chill the fuck out bro

KoraggKnightWolf:
In fact +b ~q:~a:* will never work as we don't process wildcards in ~a.  <--- Unreal already has +M to make it so that non-authed users cannot speak without the problem they cannot change their nicks ;) 
https://www.unrealircd.org/docs/Channel_Modes <--- if you want to call others out with documentation do please make sure to read more than just one link ;)

KoraggKnightWolf:
Seems like you are just on a crusade to rip out the ability to disallow nick changes entirely more than fixing a race condition in services when regaining a nick...

ddevault:
And it seems like you're a random idiot who wandered into the thread to spew garbage for attention

KoraggKnightWolf:
lmao you wanna call me an idiot when you called someone else cancer in another thread cause they disagreed, Unreal has +M use that if you want to disallow non-authed users from speaking and not affecting their usage of NICK

ddevault:
Oh give it a rest you fucking twerp, other people have already said as much and you're contributing nothing of value to this already-resolved discussion. Keep masturbating in this thread if you want, but I'm unsubscribing so I don't have to read your bullshit.

KoraggKnightWolf:
please delete yourself yes do that, and awwww you gotta throw insults cause you got cornered, just another lil wannabe skid ^^ seen enough of you scrubs, if you wanna troll, do that elsewhere and learn some better insults, cheers!

PR #119, opened 2020-08-24
Baltics nickchars support

moseslecce:
Allow support for Baltic languages in the nickname characters (Latvian, Estonian and Lithuanian) when using utf8.  These characters exist under the Latin Extended-A character set and can be used with Latin-utf8.

### DIFF
--- a/src/modules/charsys.c
+++ b/src/modules/charsys.c
@@ -85,6 +85,7 @@ struct LangList
 /* MUST be alphabetized (first column) */
 static LangList langlist[] = {
 /*	{ "arabic",       "ara", LANGAV_ASCII|LANGAV_ISO8859_6 }, -- TODO: check if this has issues first! */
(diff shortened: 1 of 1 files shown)


syzop:
Thanks for the contribution!
Would it be possible to split them up in the 3 languages, like the rest of the system? (latvian-utf8, estonian-utf8, lithuanian-utf8)
Also, you can use LANGAV_LATIN_UTF8 for all this (no need for a new LANGAV_LATIN_EXT_A_UTF8). I know it's a tad confusing, but the system is to avoid clashes in codepages and mixing scripts. In UTF8 we only care about the script language so then the baltic languages fall under the "latin script in utf8" group, doesn't matter if they are in the extended block or not.

With those changes, I think it should be fine to commit.

moseslecce:
No problem glad to help and contribute.  We've now made the changes you have requested and tested it locally and it does work as expected.

syzop:
Great, thanks for the nice patch, the PR has now been merged. The new functionality will be in next UnrealIRCd 5 release.

moseslecce:
You're very welcome.

PR #120, opened 2020-09-07
Bypass +H chanmode on BOTS (+B)

westor7:
I believe that when a bot joins in the channel and +H channel mode exists it should bypass it because it is useless to having history on join for bots, one issue can be if there was some bot commands executed before the bot join (e.g: !website) then when the bot joins it re-triggers that event.

Also bots can use `/HISTORY` command anytime if they want to view the channel history.

### DIFF
--- a/src/modules/chanmodes/history.c
+++ b/src/modules/chanmodes/history.c
@@ -534,7 +534,8 @@ int history_join(Client *client, Channel *channel, MessageTag *mtags, char *parv
 	if (!HistoryEnabled(channel))
 		return 0;
 
-	if (MyUser(client))
+	/* Check also if the client is a bot (+B) */
+	if (MyUser(client) && !IsBot(client))
 	{
 		HistoryFilter filter;
 		memset(&filter, 0, sizeof(filter));


jobe1986:
The server-time capability can be a work around for this by simply having the bot ignore messages with a timestamp earlier then the time the bot knows it joined the channel.

However personally I'd rather have a separate user mode that suppresses channel history playback on join that could be used by anyone, not just bots.

westor7:
This may not work if `server-time` module is not loaded, i know some networks that doesn't use `server-time` because it having some issues with kiwiirc. so i chose that way because the bot can either use `/HISTORY` command if it is script to execute that after bot join on the channel, but bypassing by default if they having +B usermode seems having a big point on exist bots.

Also yes maybe having +Y umode to display (or not) the channel history on join is not a bad idea but for now i think that way is a temporary fix because having bots and +H on the channel makes bot spam on its join.

GottemHams:
Simply ignoring history when the user is marked as a bot is too extreme and probably unclear for the owner as well ("hey why isn't it getting any history (anymore)?"). For example, I have a bot that logs channel messages and people can use it to find old messages. By simply ignoring all bots it would not receive any history to add to the database in case the bot has to reconnect, meaning those messages wouldn't be logged at all. I could have it do `HISTORY` "manually" but if `+H` is already set, the values it uses are tailored to that channel and are much better than say `/history #chan 1000` and having to check all those messages for ones we managed to store before the bot died. Even 100 messages could be too much, I know a few channels where 10 messages over __a couple days__ is a lot. ;]

Then again, I doubt many bot owners even know of the existence of `+B` so your proposal wouldn't even fly in many cases. I'm actually not even using `+B` myself as it doesn't change any functionality and I can live without the extra whois line, so why bother. :>

If the bot owner can't have it add an additional umode for itself (if we were to go with that), the bot is likely unmaintained anyways and you should probably wonder if you want it on your network still. ;]

But you never know, perhaps @syzop has a different opinion.

westor7:
As you said you already have the option to code on your bot to manually get the channel history via `/HISTORY` after the join but having a simple bot to get the channel history without having a proper way to stop this it will cause the bot to re-trigger his codes even in simple bots for non-expert users, the overall right now is that there is a way to retrieve the history of a channel but there is not a way to stop getting it automatically when you join on the channel and this can make troubles in some cases.

Also i cannot see that as extreme LOL! ok a new umode like `+Y` would solve this in all ways, but i think creating a new umode just for history is too much. :P

westor7:
Bump!

What about a new configurable setting like **set::history::channel::playback-on-join::disable-on-bots**

Example:

```
set {
        history {
                channel {
                        /* How many lines to playback on join? */
                        playback-on-join {
                                disable-on-bots yes; // <--------------------
                                lines 15;
                                time 1d;
                        }
                        /* How much history to keep. These are the
                         * upper maximums for channel mode +H lines:time
                         */
                        max-storage-per-channel {
                                lines 200;
                                time 1w;
                        }
                }
        }
}
```

syzop:
I agree with Gottem's comment https://github.com/unrealircd/unrealircd/pull/120#issuecomment-688410303.
As for the disable-on-bots option.. something like that would be ok, yeah.
Let's close this PR. Better use the bug tracker for any discussion.

PR #123, opened 2020-10-11
Fix when +f to not kick if is already banned

westor7:
Currently when someone is banned or muted (~q extban) then if he keeps flooding on the channel then the +f triggers this action and kick the user away from the channel, as it should not and should only appear an error message that he cannot send to the channel because he is banned.

Example Image: 

image

(Not sure if this change is correct, but i gave a try :P)

- Thanks!


syzop:
Hmm.. I don't really like the performance effects of this, it means we call is_banned() twice on every channel message on +f channels.

And it's kinda a corner case anyway... should we really worry about it?
If you find it important enough, then please report it on https://bugs.unrealircd.org/
Since it won't be as simple as this PR, I am closing the PR.

PR #124, opened 2020-10-18
Add $(DESTDIR) support for 'make install'

robert-scheck:
When packaging UnrealIRCd as RPM, `make install` needs to install the files into `$RPM_BUILD_ROOT` rather into `/`. Just changing the paths via `./Config` or `./configure` does not fit, because otherwise UnrealIRCd is finally looking for `$RPM_BUILD_ROOT/etc/unrealircd/` rather `/etc/unrealircd/`. It's fully backwards-compatible, because normally `$DESTDIR` is not being passed.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -164,71 +164,71 @@ depend:
 	done
 
 install: all
-	$(INSTALL) -m 0700 -d @BINDIR@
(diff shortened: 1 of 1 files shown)


syzop:
Looks fine to me, so merged, thanks!

PR #125, opened 2020-10-19
Exit with a successful return code upon receipt of SIGTERM

robert-scheck:
Return code 255 (like any non-zero return code) indicates a failure on POSIX systems. Actually systemd.unit gets upset about this wrong exit code when stopping unrealircd.service systemd unit.


syzop:
Sure, fine by me, merged :)

PR #129, opened 2021-01-17
Move account check to one place to avoid accepting a temporary "numeric" account name as a valid login.

k4bek4be:


syzop:
Thanks for the report and the suggestions!
Applied something similar now through 3 commits: https://github.com/unrealircd/unrealircd/commit/06c0a34ab14d197ebae48603332523fd72dfae31 https://github.com/unrealircd/unrealircd/commit/68d172854dfe2d5c6dd35f271327bead89f6deea https://github.com/unrealircd/unrealircd/commit/2afc57aa38f7d78759b646ccb1f9fc2f9ea4cc85

PR #132, opened 2021-02-01
Update help.conf

TehPeGaSuS:
**ELINE** and **[G]ZLINE** commands do accept `nick` as a parameter when setting.


TehPeGaSuS:
Need to rework

PR #135, opened 2021-05-05
Add a module to support the +draft/display-name client tag

jwheare:
https://bugs.unrealircd.org/view.php?id=5867

This is a modified copy/paste of `typing-indicator.c`, without the whitelisting of tag values in `ti_mtag_is_ok`

https://github.com/ircv3/ircv3-specifications/pull/452 (PR not yet merged)

I appreciate you probably don't want this merged into unrealircd until the spec is at least merged to IRCv3 in draft state. But I'm putting it here now because I don't foresee any changes that would affect this module. And I've made it now so may as well.


Shillos:
Hope to see that here and on irccloud!  :-)

syzop:
Would need some more thinking, changes and discussion in https://bugs.unrealircd.org/view.php?id=5867 and it seems there is some specification polishing needing to do at https://github.com/ircv3/ircv3-specifications/pull/452 as well (that alone is not a problem per se, by the way).
I think in the meantime it is best to close this PR than to keep it lingering. I'm sure it can be used for inspiration when if it is added though. Thanks!

PR #137, opened 2021-05-29
Unreal42

darksis:
please add modules to unrealircd42  Blocks clients not replying to CTCP like botnets/spambots/floodbots


syzop:
Clearly a mistake.

PR #140, opened 2021-06-17
Allowing multiple masks in "deny link" blocks.

Polsaker:
The goal of this patch is allowing the following syntax for `deny link` blocks:

```
deny link {
	mask {
		*;
		!foobar.hub;
	};
	rule !connected(foobar.hub);
	type all;
};
```

to prevent the current server from accepting connections from other leaf servers while the server is separated to the network (to prevent leaves with multiple autoconnects from bridging between two hubs)


syzop:
Thanks for the suggestion. And perfectly done patch too!

PR #141, opened 2021-06-18
Fix mask matching for `deny link::mask` blocks

Polsaker:
Small slip up from my previous PR.

For our use-case of `deny link::mask` it really makes more sense to deny linking when *no masks* match.

Take this block for example:

```
deny link {
	mask {
		!foobar.hub;
                !somethingelse.hub;
	};
	rule !connected(foobar.hub)&&!connected(somethingelse.hub);
	type all;
};
```

would end up denying all links because you can't match `foobar.hub` and `somethingelse.hub` at the same time. This is fixed by having the function return 0 on the first non-match

### DIFF
--- a/src/misc.c
+++ b/src/misc.c
@@ -1166,7 +1166,9 @@ int unreal_mask_match(Client *client, ConfigItem_mask *m)
 	return 0;
 }
 
-/** Check if a string matches any of the masks in the mask list */
+/** Check if a string matches any of the masks in the mask list.
+ * Returns 1 if no masks match the string
+ */
 int unreal_mask_match_string(const char *name, ConfigItem_mask *m)
 {
 	for (; m; m = m->next)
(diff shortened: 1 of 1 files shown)


syzop:
This patch would break something like ```mask { *.com; *.net; }``` where a server is called ```irc.example.net``` and the ```return 0``` happens because it does not match ```*.com```.
I also noticed another bug in the previous patch. You use a ```m->mask+1``` even if there is no ```!```, thus skipping the first letter of a server for normal matching.

I am closing this PR and did two more commits myself, last important one being https://github.com/unrealircd/unrealircd/commit/f71f6c8fe3726862fdfe96304927ac0134d905f2. The new rules are explained in that commit but in short they provide the functionality you wanted. I will update the docs later.

I think/hope this is enough fiddling with that code for now :D

PR #142, opened 2021-06-23
Don't handle certfp in 'MD client'

stefansaraev:
certfp may be used in oper{} blocks and sasl, allowing other
servers to spoof it makes certfp useless for auth.

### DIFF
--- a/src/modules/md.c
+++ b/src/modules/md.c
@@ -101,6 +101,10 @@ CMD_FUNC(cmd_md)
 
 	if (!strcmp(type, "client"))
 	{
+		// dont allow certfp spoofing
+		if (!strcmp(varname, "certfp"))
+			return;
+
 		Client *target = find_client(objname, NULL);
 		md = findmoddata_byname(varname, MODDATATYPE_CLIENT);
 		if (!md || !md->unserialize || !target)


syzop:
Bit of an edge case here, you are basically saying that a rogue server can cause someones oper block (or other) to match even though it should not.

Servers are considered to be trusted by UnrealIRCd, so if you cannot trust them anymore then you are fighting a loosing battle. The privilege escalation to ircop that you suggested that is possible via this 'bug', and after this would be 'fixed', could just as easily be done if the same rogue server introduced a remote IRCOp and started executing whatever ircop-command you wanted to be executed in the first place, since remote IRCOps are just as trusted as local IRCOps.

Similarly, if you think a remote server spoofing authentication is an issue, then you have issues with SASL by certfp too? After all, any server (in between) could send a fake certificate fingerprint to services to authenticate the user, or the other way around: a server in-between could just pretend that services authenticated the user (SVSLOGIN or SVS2MODE +d), and so on.
Like I said, you are in a loosing battle there if you cannot trust servers anymore.

Now, all that being said, it makes no sense for a remote server to set a "certfp" of one of our locally connected users. There is no use case or justification for that, so you do have a point there. I have now updated the module API to block writes to clients that are our own by default (such as "certfp" but also "webirc" and such), and allow modules to opt-in (for example k4be will probably do this for "geoip").

stefansaraev:
thanks! :)

stefansaraev:
@syzop fyi.

[Fri Jun 25 13:35:39 2021] - Remote server 'irc.freenode.ceo' tried to write moddata 'link-security' of a client from ours 'irc.freenode.ceo' -- attempt blocked.

ooops. it seems this will breal link-security downgrade warnings.

it may be good idea to keep "some" metadata always writable.

syzop:
Indeed, thanks, https://github.com/unrealircd/unrealircd/pull/145 applied.

PR #143, opened 2021-06-24
Allowing sending cookies/tokens in remote PINGs

Polsaker:
This allows sending a cookie as a third parameter to the `PING` command to remote servers, useful to calculate the RTT across a whole network.

Example:

```
<< PING * some-server.com <cookie>
>> PONG some-server.com mynick <cookie>
```

### DIFF
--- a/src/modules/pingpong.c
+++ b/src/modules/pingpong.c
@@ -62,11 +62,12 @@ MOD_UNLOAD()
 ** cmd_ping
 **	parv[1] = origin
 **	parv[2] = destination
+**	parv[3] = (optional) cookie
 */
 CMD_FUNC(cmd_ping)
 {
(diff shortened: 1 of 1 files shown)


syzop:
I suppose we could do that, but... 

Just wanted to say that you don't need a cookie to keep track of requests. You can use labeled-response for that:
```
CAP REQ :message-tags labeled-response batch
@label=zzz PING testlink.test.net
@label=zzz;msgid=QlpJ16QJkgsAzv3iMXOfvc;time=2021-06-25T12:54:05.000Z :testlink2.test.net PONG testlink2.test.net :testlink.test.net
```

Polsaker:
I... totally didn't consider that :D

Thanks!

PR #144, opened 2021-06-24
Designate end of HELP(OP) output

ValwareIRC:
repurpose unused numeric response RPL_HELPHLP to signify end of output for use cases such like client-side scriptables (viewing) and in case a bot gains self-awareness :> ♥

### DIFF
--- a/src/modules/help.c
+++ b/src/modules/help.c
@@ -56,6 +56,7 @@ MOD_UNLOAD()
 
 #define HDR(str) sendto_one(client, NULL, ":%s 290 %s :%s", me.name, client->name, str);
 #define SND(str) sendto_one(client, NULL, ":%s 292 %s :%s", me.name, client->name, str);
+#define HND(str) sendto_one(client, NULL, ":%s 292 %s :%s", me.name, client->name, str);
 
 ConfigItem_help *find_Help(char *command)
 {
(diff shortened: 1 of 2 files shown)


fkm:
My recommendation for the numeric names:
* 290 RPL_HELPHDR
* 292 RPL_HELPTLR
* 293 RPL_HELPHLP

My reference is: https://www.alien.net.au/irc/irc2numerics.html

It's old and potentially wrong. But there seems to be a consensus when searching for the numbers on Google. They probably all just copied the names from one source at some point. But it's dangerous to go against established standards. Even if they are not official.

But maybe someone from the team has some input to that.

syzop:
End-users haven't been missing this numeric, since they have been living without it for 20 years and.. well.. there's not much use for them. But indeed, you mention bots or automated things, eg one can imagine throwing it in some kind of GUI popup or something...
In the past I may have been OK with this, but nowadays we have ```labeled-response```. It basically makes adding any new "end" numerics unneeded as it is redundant information. Also, numerics are a scarce resource, so if there's no good reason to add one, then don't add one in my opinion (in that sense, my opinion deviates from several other ircd brands who tend to add lots of numerics, usually to get rid of notices, which is also a valid point of view).

It's not like I have a strong opinion on the matter but I'm inclined to reject it. Unless the other devs have a different opinion?

PR #146, opened 2021-06-27
Sending back the labeled-response for PING commands.

Polsaker:
Making labeled-response work for `PING`, so we can have a proper alternative to #143


syzop:
I think this fix may have unwanted side-effects (not for PING/PONG), which is why I will look at it later again and have not merged it yet. Just so you know.

Polsaker:
re targeted to unreal 6 on #170

PR #148, opened 2021-07-02
cap: Remove trailing whitespaces from CAP ACK/NAK

progval:
I don't see any reason for them to be there, and it's unclear if
<https://ircv3.net/specs/extensions/capability-negotiation.html>
allows it (it just says 'space-separated list of capabilities')

### DIFF
--- a/src/modules/cap.c
+++ b/src/modules/cap.c
@@ -300,6 +300,7 @@ static void cap_req(Client *client, const char *arg)
 		 */
 		if (buflen + plen + strlen(cap->name) + 6 >= BUFSIZE)
 		{
+			pbuf[0][plen-1] = '\0'; // Remove trailing space
 			pbuf[1][0] = '\0';
 			plen = 0;
 			i = 1;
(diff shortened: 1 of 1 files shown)


syzop:
If an IRC client has issues with (such) trailing whitespace then that client should really be fixed.

Not the same, and perhaps a confusing example, but it is a bit like clients who fail on `x y z` being sent by `x y :z` by some IRCds. As long as `z` contains no spaces, the client should accept it as the same. If a client fails on that, eg makes a difference between with the `:` and without it, it means they are parsing at a too low level or parsing with insufficient checks that should be common when programming.

Now, we still allow patches like these in sometimes, eg for aesthetic reasons. For this particular patch, I don't like the direct accessing of the character array, which is apparently common in this code. I would prefer to rewrite all the code in `cap_req()` to conform to the new style of doing things in a safe & clean way instead of the old focus on performance. That is just a low to normal priority item for me (although higher prio than "removing trailing whitespace"), and right now we are in a change freeze.

I think it's best for me to just close this PR and we'll see one day if I rewrite the thing there in the next few years. Then I am likely to think of stripping the trailing whitespace as well. I think I will remember, but if not then you are free to prod me after the rewrite.

PR #149, opened 2021-07-02
Add support for +draft/react and load the module by default.

progval:
See https://ircv3.net/specs/client-tags/reply for the draft.

This is copy-pasted from src/modules/reply-tag.c minus the
(lack of) validation of the value.

### DIFF
--- a/doc/conf/modules.default.conf
+++ b/doc/conf/modules.default.conf
@@ -208,6 +208,7 @@ loadmodule "labeled-response"; /* correlate requests and responses easily */
 loadmodule "bot-tag"; /* indicate the message comes from a bot (draft/bot) */
 loadmodule "typing-indicator"; /* typing indicator in PM and channels (+typing) */
 loadmodule "reply-tag"; /* indicate to which message you are responding (+draft/reply) */
+loadmodule "react-tag"; /* indicate reactions to other messages (+draft/react) */
 loadmodule "clienttagdeny"; /* informs clients about supported client-only message tags */
 loadmodule "sts"; /* strict transport policy (set::tls::sts-policy) */
 loadmodule "link-security"; /* link-security announce */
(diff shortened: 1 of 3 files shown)


syzop:
I'm afraid I have no plans to accept this in UnrealIRCd. My main problem with this is that it means *some* clients will see reactions and others will not see them at all. That is highly confusing and destroying the IRC experience we have had for a long time where all clients see the same discussion. I think that's really bad for IRC.
I'm not going to redo the entire discussion about this, which is why I will post a long answer below, and you can also read https://github.com/ircv3/ircv3-specifications/issues/333.

At UnrealIRCd we have always been progressive about IRC. We have been accused by specification- and protocol-purists for violating the RFC for decades. Something we always did and will keep doing to advance IRC, long before IRCv3 existed. Nowadays a majority considers that good and modern.
With IRCv3 we continue to pick good features that bring IRC forward. It's fine if some clients see things slightly different than others, eg cap chghost versus rejoin on host change, or the typing tags or reply tag.. that's all fine, it's not really at the expense of others. But with this feature it is an issue. It breaks with the IRC tradition where everyone can follow the same conversation. With this feature you are basically locked out of the conversation. If you don't support the +react feature then you don't see part of the conversation. It's a moderation problem but also goes far beyond moderation. It's not good and like I said it destroys the IRC experience when not everyone sees the same discussion. 
Now, this may not have that much of an impact if the tag is rarely used, but then one could just as well not allow the tag at all.. like in our case.

Also, and I know the two features are not identical, but +reply can be used in many of the situations where one would use +react. The good thing is that the +reply tag does not have the downsides that the +react tag has. With +reply, users who are on IRC clients that do not support the +reply tag can still see the content of the messages/reactions from +reply users. We do support +reply for that reason.

You are still free to submit it as a 3rd party module at unrealircd-contrib, if you wish. The [rules for unrealircd-contrib](https://www.unrealircd.org/docs/Rules_for_3rd_party_modules_in_unrealircd-contrib) are far more lax.

PR #150, opened 2021-07-03
chathistory: Use more explicit messages on INVALID_TARGET failure message

progval:


src/modules/chathistory.c
```diff
@@ -186,9 +186,16 @@ CMD_FUNC(cmd_chathistory)
 	}
 
 	channel = find_channel(parv[2], NULL);
-	if (!channel || !IsMember(client, channel) || !has_channel_mode(channel, 'H'))
+	if (!channel || !IsMember(client, channel))
```
k4bek4be: Maybe also handle the third case: the channel exists, but client is not its member?

progval: Indeed, done.

PR #151, opened 2021-07-05
Adding `sequential-fallback` autoconnect strategy

Polsaker:
This new autoconnect strategy is identical to the `sequential` one, but it will re-try the first node upon a successful connection.

This is useful if you have a primary hub and all your other autoconnects are less desirable, fallback links.


syzop:
This is a good idea, but now I wonder if maybe this should be the default, which is why I didn't immediately hit the "merge" button just yet :D. Will get back on this...

syzop:
Alright, I'm just going to merge this. We can look at making it the new default in a later version. And thanks again! :D

PR #152, opened 2021-07-09
Implementing IPv6 -> IPv4 fallback

Polsaker:
This PR will make unreal attempt to link over IPv4 when getting timeout or network unreachable errors over IPv6, as opposed to just failing to connect.


syzop:
I don't want to touch 5.x for this, as most people did not have an issue with this. Presumably they are used to using an IP address in `link::outgoing::hostname` and can live with this effect.

For 6.x (too late in the development process, after the freeze) I looked at [Happy Eyeballs v2](https://datatracker.ietf.org/doc/html/rfc8305) as I would want to "do things good, once and for all". We could do it as they propose there or with a slight variation like this:
- Start an IPv6 lookup
- If we have an IPv6 record, start connecting there
- Upon IPv6 DNS lookup response (both found and not found) or after a DNS lookup timeout: do the IPv4 lookup
- If we have an IPv4 record, and the IPv6 is not connected yet, start connecting too
- Then whichever connection is established wins and the other one is cancelled.
And with "established" I mean SYN-ACK, with zero IRC protocol messages have been sent yet.

Of course, that would require a more major overhaul of things, so.. I don't know how realistic it is to do without risking bugs or breakage (multiple objects referring to the same link block, ref counts, and so on). We can try, though.

As this PR won't make it into unreal52, I am closing it. I copy-pasted the above to the bug issue tracking this https://bugs.unrealircd.org/view.php?id=4687

PR #155, opened 2021-07-10
Fix for mod data client modification restriction inconsistency

jobe1986:
Allow a server to modify mod data for its own clients and itself even if it is a client of the server receiving the MD message. This should fix bug https://bugs.unrealircd.org/view.php?id=5960


jobe1986:
No longer needed

PR #156, opened 2021-07-10
Make CHATHISTORY not FAIL on a request for -H channel, sending empty history instead

k4bek4be:


syzop:
Some say current behavior is totally fine and how it should be, others (may) want something else. Have asked again today upstream if this is what they want/need.

Pending clear up there...

PR #158, opened 2021-07-11
Removing seemingly needless looping on SQUITs.

Polsaker:
All this recursing seemed to be a leftover from ancient times and
was degrading the performance.

Hoping Syzop can provide more insight on this :D

### DIFF
--- a/src/misc.c
+++ b/src/misc.c
@@ -567,28 +567,6 @@ void remove_dcc_references(Client *client)
 }
 
 /*
- * Recursively send QUITs and SQUITs for cptr and all of it's dependent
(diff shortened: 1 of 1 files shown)


tzvetkoff:
I assume this was only needed for servers not supporting `NOQUIT` cap.
Since nowadays pretty much everything supports it, well... let's move forward!

syzop:
Thanks. I am not too comfortable doing this in 5.x but have done now 6.0.0, will be in 6.0.0-beta2 whenever it is released.
Commit: https://github.com/unrealircd/unrealircd/commit/1b308c7ca05308c815a0e3f1e203e47711ae6457

PR #159, opened 2021-07-11
Sending WALLOPS back to the oper that issued it.

Polsaker:
This reassures the oper that the message was actually sent, and
makes it consistent with the behavior of GLOBOPS and NOTICE $*

### DIFF
--- a/src/modules/wallops.c
+++ b/src/modules/wallops.c
@@ -73,5 +73,8 @@ CMD_FUNC(cmd_wallops)
 		return;
 	}
 
+	if (MyUser(client))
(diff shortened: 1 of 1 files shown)


syzop:
Yeah this is basically the same as it ended up in UnrealIRCd 6 (9f56fb4aa758df814a390812472ca0760d7cc3c7) after discussion on https://github.com/unrealircd/unrealircd/pull/171 and https://bugs.unrealircd.org/view.php?id=5764

I didn't notice the commit and this patch is pretty much identical, thanks! :D

PR #163, opened 2021-08-14
Send RPL_WHOISACTUALLY in addition to RPL_WHOISHOST

progval:
Unlike RPL_WHOISHOST, it contains the user@host and ip as separate parameters,
instead of having them in the free-text parameter.

To my knowledge, there are three different syntaxes for RPL_WHOISACTUALLY:

1. `<client> <nick> :is actually ...` (bahamut)
2. `<client> <nick> <host> :Is actually using host` (charybdis family)
3. `<client> <nick> <username>@<host> <ip> :Is actually using host`
   (ircu2, ergo, hybrid, ...)

I chose the third one because it contains the most info, and is the most
common one.

I kept RPL_WHOISHOST despite being redundant for backward compatibility,
but I can remove it if you prefer.

### DIFF
--- a/include/numeric.h
+++ b/include/numeric.h
@@ -193,6 +193,7 @@
 
 #define RPL_INVITELIST       336
 #define RPL_ENDOFINVITELIST  337
+#define RPL_WHOISACTUALLY    338
 
 #define RPL_LISTSYNTAX       334
 #define RPL_WHOISBOT	     335
--- a/src/modules/whois.c
+++ b/src/modules/whois.c
@@ -136,6 +136,12 @@ CMD_FUNC(cmd_whois)
 			}
 			if ((target == client) || IsOper(client))
 			{
+				/* Clients should prefer RPL_WHOISACTUALLY as it is easier to parse,
+				 * but we provide RPL_WHOISHOST (despite it being redundant)
+				 * for backward compatibility */
+				sendnumeric(client, RPL_WHOISACTUALLY, target->name,
+					(MyConnect(target) && strcmp(target->ident, "unknown")) ? target->ident : "*",
+					target->user->realhost, target->ip ? target->ip : "");
 				sendnumeric(client, RPL_WHOISHOST, target->name,
 					(MyConnect(target) && strcmp(target->ident, "unknown")) ? target->ident : "*",
 					target->user->realhost, target->ip ? target->ip : "");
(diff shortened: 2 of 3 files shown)


progval:
FYI, I sent [a similar PR](https://github.com/inspircd/inspircd/pull/1923) to InspIRCd, but they decided to replace RPL_WHOISHOST with RPL_WHOISACTUALLY instead of sending both

syzop:
I'm not considering this for UnrealIRCd 5.x.

I have looked at this during UnrealIRCd 6 development, but it was behind closed doors at the time so I could not comment about the results here. I agree with InspIRCd that *if* you would go this route then you should send one of the numerics and not both. The problem is it caused display issues in /WHOIS with the first client that I tried, which was mIRC, or at least half of the information is missing and it looks weird too. This made me wonder how common this numeric really is (or at least in the variant you proposed), if such a common client does not display it correctly. So I did not put it in UnrealIRCd 6. After that, I simply forgot about it.

In general I don't mind changing things in the interest of standardization, it makes life more simple for (future) client coders. However, the benefit to end-users of it is limited (in this case they would see the same information, just in a different way, or actually in this case possibly a near-identical way). So, with the *benefit* being limited, as soon as it causes any display issues (a *cost*), the *benefit/cost ratio* is not good anymore, which makes me lose interest.

So, how to move forward? If you are passionate about this particular thing, then feel free to try with a good amount of major clients: see if there is loss of information or other issues in the /WHOIS display if you use RPL_WHOISACTUALLY instead of RPL_WHOISHOST with the patch you proposed, and then hunt the client developers down if such display issues would occur.
If you can then report back here with a list of clients and the result of that looks good (eg have them fixed or did not have them in the first place) then that would help with acceptance of this numeric change. And yes, this could also involve us waiting a while until clients have fixed things and giving end-users some time to catch up on those new versions, as I see no reason to rush this.

syzop:
Closing this PR, since it will need to be retargeted for U6 (taking into account the notes of above).

progval:
> However, the benefit to end-users of it is limited (in this case they would see the same information, just in a different way, or actually in this case possibly a near-identical way).

Another benefit is that RPL_WHOISACTUALLY is parseable, either by bots or clients wishing to implement more "fancy" displays; while RPL_WHOISHOST can't reliably be displayed as anything but free-form text.

PR #164, opened 2021-09-01
I know it's niche but: be more inclusive of 'all-admin' 

ValwareIRC:
for all-admin setups (yes they exist)
removed the warning and allow the mass setting/adding of channelmodes 'vhoaq' via SVSMODE


ValwareIRC:
I genuinely have no idea how this messed up so badly

PR #166, opened 2021-09-04
introduce service tag

ValwareIRC:
this allows clients to properly ensure the client requesting a password is not someone holding the nick hostage after a netsplit case or something.

### DIFF
--- a/src/modules/service-tag.c
+++ b/src/modules/service-tag.c
@@ -86,7 +86,7 @@ void mtag_add_servicetag(Client *client, MessageTag *recv_mtags, MessageTag **mt
 	if (IsUser(client) && has_user_mode(client, 'S'))
 	{
 		MessageTag *m = safe_alloc(sizeof(MessageTag));
-		safe_strdup(m->name, "unrealircd.org/service");
+		safe_strdup(m->name, "inspircd.org/service");
 		m->value = NULL;
 		AddListItem(m, *mtag_list);
 	}
--- a/src/modules/service-tag.c
+++ b/src/modules/service-tag.c
@@ -49,7 +49,7 @@ MOD_INIT()
 	MARK_AS_OFFICIAL_MODULE(modinfo);
 
 	memset(&mtag, 0, sizeof(mtag));
-	mtag.name = "unrealircd.org/service";
+	mtag.name = "inspircd.org/service";
 	mtag.is_ok = servicetag_mtag_is_ok;
 	mtag.flags = MTAG_HANDLER_FLAGS_NO_CAP_NEEDED;
 	MessageTagHandlerAdd(modinfo->handle, &mtag);
--- a/src/modules/service-tag.c
+++ b/src/modules/service-tag.c
@@ -1,0 +1,93 @@
+/*
+ *   IRC - Internet Relay Chat, src/modules/bot-tag.c
+ *   (C) 2021 Syzop & The UnrealIRCd Team
+ *
+ *   See file AUTHORS in IRC package for additional names of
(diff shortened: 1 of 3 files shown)


src/modules/service-tag.c
```diff
@@ -0,0 +1,93 @@
+/*
+ *   IRC - Internet Relay Chat, src/modules/bot-tag.c
```
progval: `src/modules/service-tag.c`

src/modules/service-tag.c
```diff
@@ -0,0 +1,93 @@
...
+void mtag_add_servicetag(Client *client, MessageTag *recv_mtags, MessageTag **mtag_list, char *signature);
+
+MOD_INIT()
+{
+	MessageTagHandlerInfo mtag;
+
+	MARK_AS_OFFICIAL_MODULE(modinfo);
+
+	memset(&mtag, 0, sizeof(mtag));
+	mtag.name = "unrealircd.org/service";
```
progval: Why not `inspircd.org/service`?

ValwareIRC: Why inspircd? It is UnrealIRCd.
Just wondering?

progval: `inspircd.org/` is used as a unique namespace, it doesn't mean only InspIRCd can use it. As long as Unreal behaves like InspIRCd, there's no reason to use a different name.

ValwareIRC: I mean, I only found out inspircd had it today because I asked after I had the idea, I didn't take the idea from there, I don't see why it shouldn't use the vendor name

progval: So client developers don't have to track all implementations of the same concept, while making sure they are not subtly different.

ValwareIRC: I agree in a sense, but I think here, the concept may have already been implemented in another ircd, however, to me, the 'concept' is "/service" and I think the vendor should be described.

And client implementations, the client would most likely have tokenized the message tags (as I do when client-side scripting), it's very easy to tokenise a token to see if "/service" is part of the token, and describing the vendor to me seems like additional security should a client choose to pursue that.

emersion: I agree with @progval, if you're implementing the same functionality as inspircd I'd much rather see you use the existing cap. An implementation using the vendored cap from a different vendor isn't unheard of.

The string matching with "/service" is unsafe, a different vendor could choose to give a completely different meaning to a tag containing this string.

progval: The vendor is [used as a namespace](https://ircv3.net/specs/extensions/message-tags#vendor-specific) to avoid conflicting names when two implementations use the same word for different concepts. (to avoid the same clusterfuck as conflicting modes, like `+q`).

If clients ignore the vendor part and only use the name after the slash, this nullifies the usefulness namespacing; we might as well not use vendored names at all.

Additionally, splitting on `/` is forbidden by the message-tags spec: ["Implementations MUST treat tag key names as case-sensitive opaque identifiers"](https://ircv3.net/specs/extensions/message-tags#format)

ValwareIRC: Well why didn't you say ;D

progval: You forgot to update this one (line 52)

ValwareIRC: fixed lol 👀👀

syzop:
I don't find this a useful feature. Any decent network prevents normal users from taking the nick NickServ and ChanServ, we even have it in the example.conf by default. That is way more useful/important than adding a tag.

You are free to code a 3rd party module for it, of course.

PR #168, opened 2021-10-03
Add CentOS 7 to Release Notes

robert-scheck:
CentOS 7 ships OpenSSL 1.0.2k


syzop:
We could have added it, but.. this change would not have changed the release notes of the already released version. I think it's OK to close by now.
Thanks for maintaining RPM packages btw, I've never used it myself but I'm sure it is appreciated by others :)

PR #170, opened 2021-10-30
Sending labeled-response on remote TSCTL and PING replies

Polsaker:


syzop:
As I mentioned in the previous PR my gut feeling says this fix is wrong as it has the potential to cause issues in other areas. But... I have not had the chance to look into this yet, eg that issue and/or an alternative patch that would work as well.
It is good you brought it up again, though, and I will look at it for 6.0.1/6.0.2/.. as it would be nice to have this functionality working. During UnrealIRCd 6 development I focused on other (major) areas and did not go through all PR's / other bugs/feature enhancements in the bug tracker "that would also be nice" since it would have delayed the release too much.

Just saying this, in case you wonder why it is lingering.

PR #171, opened 2021-10-30
Sending wallops back to the user that issued it

Polsaker:
Small QoL improvement so that we can be 100% sure that the wallops was actually sent (it also now behaves like globops)

### DIFF
--- a/src/modules/usermodes/wallops.c
+++ b/src/modules/usermodes/wallops.c
@@ -63,7 +63,7 @@ MOD_UNLOAD()
  * @param pattern	The format string / pattern to use.
  * @param ...		Format string parameters.
  */
-void sendto_wallops(Client *one, Client *from, FORMAT_STRING(const char *pattern), ...)
+void sendto_wallops(Client *from, FORMAT_STRING(const char *pattern), ...)
 {
 	va_list vl;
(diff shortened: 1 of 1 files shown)


syzop:
Since you removed _all_ the skipping over, in a multi-server scenario, wouldn't this cause a receiving server to send the message back to the origin server? Causing a never ending storm of A->B, B->A, A->B, B->A, etc. ?

Polsaker:
This is not the case as far as we've tested on Pissnet (we tested 6.0 patched <-> 6.0 patched, 6.0 patched <-> 6.0 unpatched and 6.0 patched <-> 5.2)

ValwareIRC:
can't wait for this to be a thing :)

syzop:
This patch indeed sends the message back to where it came from. It doesn't cause a storm because it results in a message with a "bad direction" which is silently ignored, but I'm afraid it's still wrong.
I have now fixed it in https://github.com/unrealircd/unrealircd/commit/9f56fb4aa758df814a390812472ca0760d7cc3c7
The alternative would be to do it in the loop in sendto_wallops() at the place where it checks for == one but then that if became a bit too confusing so I like this one plain and simple :)

PR #172, opened 2021-10-31
Update short_date function to avoid potential crash on musl if year > 9999

alicetries:
The asctime() function is undefined when called with a year of >9999 (as it would generate output longer than the 26 byte buffer) according to POSIX
This only appears triggerable through sending dodgy data through S2S commands, however does cause the ircd to crash when running on a mixed musl/glibc network.


alicetries:
This may actually be better to be replaced with strftime, I'll have a bit of an investigate on that one.

PR #174, opened 2021-11-01
Update short_date function to avoid potential crash on musl if year > 9999

alicetries:
The asctime() function is undefined when called with a year of >9999 (as it would generate output longer than the 26 byte buffer) according to POSIX
This only appears triggerable through sending dodgy data through S2S commands, however does cause the musl-based ircds to crash when running on a mixed musl/glibc network.

### DIFF
--- a/src/misc.c
+++ b/src/misc.c
@@ -153,7 +153,6 @@ const char *long_date(time_t clock)
 const char *short_date(time_t ts, char *buf)
 {
 	struct tm *t = gmtime(&ts);
-	char *timestr;
 	static char retbuf[128];
 
 	if (!buf)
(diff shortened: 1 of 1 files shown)


syzop:
I think you said on IRC not to accept this PR since you would come with a better one. Or is that this one? :)
Just to clear that up...

Otherwise looks fine, as we already use `strftime()` elsewhere, and this patch gets rid of a needless buffer copy.

alicetries:
Hi Syzop,

Yep, this is the cleaner one. The original one was basically checking that tm_year <= 9999, but the current strftime approach is cleaner.

syzop:
Clean indeed. Thanks!

PR #179, opened 2021-11-03
Remove $client that is not defined there

westor7:
Officially reported by delta.

Closing https://bugs.unrealircd.org/view.php?id=5993


westor7:
Invalid.

PR #180, opened 2021-11-03
Fix on TLS errors missing the IP part

westor7:
Officially reported by delta.

Closes https://bugs.unrealircd.org/view.php?id=5993


simplexish:
After some light testing it seems that this patch fixes the issue.

syzop:
Thanks for the report and the PR. Should be fixed now.

`$client` expands to the name of the client, so nick name or server name. (any $var for something with sub-items would expand to $var.name).

In this case `$client.details` is better which can evaluate to a number of things, depending on how much info is available. In this case it will almost always evaluate to the IP address of the user.  I also changed the logging routine a bit to not show "@127.0.0.1" in this case but show it as "[127.0.0.1]" so it's clear it is (only) an IP address.

PR #182, opened 2021-11-18
Small simplification

crazycatdevs:
I guess `score += 5 + (digits - 5)` is the same than `score += digits`


syzop:
Funny indeed, makes you wonder how that happened :D

As a general principle I try to avoid touching 5.x without good reason.
Can you retarget this patch for branch `unreal60_dev`? Then I will accept it post-6.0.0, eg for 6.0.1 / 6.0.2. Thanks!

PR #183, opened 2021-11-28
Lower libmaxminddb requirement from 1.6.0 to 1.2.0

robert-scheck:
libmaxminddb 1.6.0 was released on 2021-04-29 and is not yet widely available in Linux distributions. However, the alternative GeoIP Legacy C Library will be retired at the end of May 2022 and receives currently only critical security and bug fixes. Upstream recommends to upgrade to GeoIP2 databases, which leads to libmaxminddb when using C.

See also:
 - https://github.com/maxmind/geoip-api-c/blob/main/README.md
 - https://repology.org/project/libmaxminddb/versions

### DIFF
--- a/autoconf/m4/unreal.m4
+++ b/autoconf/m4/unreal.m4
@@ -397,7 +397,7 @@ AC_DEFUN([CHECK_LIBMAXMINDDB],
 	[
 		dnl see if the system provides it
 		has_system_libmaxminddb="no"
-		PKG_CHECK_MODULES([LIBMAXMINDDB], [libmaxminddb >= 1.6.0],
+		PKG_CHECK_MODULES([LIBMAXMINDDB], [libmaxminddb >= 1.2.0],
 		                  [has_system_libmaxminddb=yes])
 		AS_IF([test "x$has_system_libmaxminddb" = "xyes"],
 		[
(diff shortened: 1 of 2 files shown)


syzop:
We use the legacy library due to licensing issues. Anyway, that is not what this PR is about:
You asked to lower the version requirement for libmaxminddb. That sounds reasonable. You propose to lower it to `>=1.2.0`. Presumably that version is picked based on the API compatibility, which.. yeah.. is a perfectly reasonable logic to follow. On the other hand, 1.2.0 was released in 2016. Lots of heap corruption and other bugs have been fixed since then. I have now changed the requirement to `>=1.4.3` in commit https://github.com/unrealircd/unrealircd/commit/a71e2051857bbd9a2039dd85daacaa734e61803a. The reason for this is that I don't want a known faulty libmaxminddb corrupt UnrealIRCd causing headaches for us when we receive crash reports with mysterious heap corruption. Version 1.4.3 (released Aug 2020) seems to fix the last known major bugs related to that. The fixes between 1.4.3 and 1.6.0 are not crucial in that sense (not even the fix in 1.5.2 related to unchecked calloc, i can live perfectly without that fix).

So, thanks for the suggestion. I hope you can live with `>=1.4.3`. If you cannot for your packaged unrealircd RPM then you would still have to patch it, but that's up to you.

PR #167, opened 2021-09-19
help: Change numerics for consistency with other IRCds

progval:
Especially Ergo, InspIRCd, and Solanum.

I'm working on a specification here: https://github.com/ircdocs/modern-irc/pull/132

PR #184, opened 2021-11-28
help: Change numerics for consistency with other IRCds

progval:
This is GH-167, rebased on unreal60_dev.

Screenshots below, with the two versions (first the current one, then the one I'm proposing)

# Looks much better

Konversation:

konversation_1.8-20121

KVIrc
kvirc_5 0 0

# Looks the same

IRCCloud:

irccloud

Srain:

srain_1.2.4

# Looks slightly worse, as the client shows the command name (here, it's `*`) before the text

Gamja:

gamja

Hexchat:

hexchat_2.14.3

irssi:

irssi_1.3

KiwiIRC:

kiwiirc

Quassel:

quassel_0.13.1

Weechat:

weechat_3.0

### DIFF
--- a/src/modules/help.c
+++ b/src/modules/help.c
@@ -54,8 +54,12 @@ MOD_UNLOAD()
 	return MOD_SUCCESS;
 }
 
-#define HDR(str) sendto_one(client, NULL, ":%s 290 %s :%s", me.name, client->name, str);
-#define SND(str) sendto_one(client, NULL, ":%s 292 %s :%s", me.name, client->name, str);
+// RPL_HELPSTART
+#define HDR(subject, str) sendto_one(client, NULL, ":%s 704 %s %s :%s", me.name, client->name, subject, str);
+// RPL_HELPTXT
+#define SND(subject, str) sendto_one(client, NULL, ":%s 705 %s %s :%s", me.name, client->name, subject, str);
+// RPL_ENDOFHELP
+#define FTR(subject, str) sendto_one(client, NULL, ":%s 706 %s %s :%s", me.name, client->name, subject, str);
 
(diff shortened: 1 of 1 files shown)


syzop:
I see this is lingering again, better to make a decision than no decision:
I find it rather ugly for those cases where it looks worse... especially since you only used `*` as a command and for something like `CHMODES` etc or any other query it would look a lot more worse... repeating that command over in over, prefixed everywhere, not nice at those clients. And... there are only two clients at which it looks better.

I am going to do what I have often done with numeric changes in the past 20 years: I am going to wait with this until clients catch up and show things more nicely. In the meantime, I don't think I should keep this PR open for the next 2 years while waiting for such a moment, that would look silly, so I am closing it. I hope you keep going on with your improvements and let's see in 2 years or so if things have evolved for the better.

progval:
If you don't mind, I'll use this PR to keep track of the progress of adding support to clients.

The Lounge already supports the new numerics better:

screenshot-2022-04-01_14-57-24

For the other clients mentioned above that need to be improved:

* https://github.com/kiwiirc/kiwiirc/pull/1509
* https://github.com/irssi/irssi/pull/1371
* https://github.com/weechat/weechat/pull/1772
* https://github.com/quassel/quassel/pull/601

syzop:
> If you don't mind, I'll use this PR to keep track of the progress of adding support to clients.

Yes, no problem :)

progval:
Four major clients (KiwiIRC, irssi, weechat, and Quassel) merged my PRs to add support for the new numerics. This leaves only Hexchat (which is abandonware) and Gamja. Is it good enough to merge this PR now?

PR #185, opened 2021-11-28
Small simplification

crazycatdevs:
cf PR #182 : 

> I guess `score += 5 + (digits - 5)` is the same than `score += digits`


syzop:
Thanks! Still don't know how it ended up like that before :D

PR #188, opened 2021-12-16
Add SOURCE_DATE_EPOCH support for build date

kpcyrd:
This uses the timestamp in [`$SOURCE_DATE_EPOCH`][1] if set and falls back to the current behavior otherwise. It tries `date -d "@..."` first and falls back to `date -r "..."` to support BSD date(1) too.

I'm not sure if I'm opening this PR to the right branch, let me know and I can rebase if needed.

Thanks!

[1]: https://reproducible-builds.org/docs/source-date-epoch/

### DIFF
--- a/src/version.c.SH
+++ b/src/version.c.SH
@@ -22,7 +22,9 @@ generation=`expr $generation + 1`
 export LANG=C
 export LC_TIME=C
 export LC_ALL=C
-creation=`date | \
+SOURCE_DATE_EPOCH="${SOURCE_DATE_EPOCH:-$(date +%s)}"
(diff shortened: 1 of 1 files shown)


syzop:
Sounds good. I assume you tested it on both Linux and FreeBSD indeed. If you could rebase for unreal60_dev that would be great.

kpcyrd:
I've rebased and updated the PR to unreal60_dev. The code is copied from the FreeBSD compatible reproducible builds example but I've only tested this on Linux because I don't have a FreeBSD system to test this with.

kpcyrd:
Thanks!

syzop:
I took the chance. Seems to work on all Linux and FreeBSD (12) buildbots (not public).

Thanks! :)

PR #190, opened 2021-12-30
Added a space-typo

westor7:
Closes https://bugs.unrealircd.org/view.php?id=6049

Reported by armyn


syzop:
Unfortunately it is more complicated than that. I screwed up earlier fixing a "space too much" issue and created this problem.

The thing is that `$destination` may exist or it may be empty, if it it exists it must output something like `[cmd: PRIVMSG victim: this is the string]` and if it does not exist it must output `[cmd: NICK: badnick]`. So sometimes you need to add a space, and sometimes not. Done now in https://github.com/unrealircd/unrealircd/commit/40c528acde5212718e93aa1a5fc90d83c9623f80.

PR #191, opened 2022-02-17
small change to let users see SVSMOTD lines

ValwareIRC:
just a tweak to allow users to see lines added by svsmotd
patch for: https://bugs.unrealircd.org/view.php?id=6070


syzop:
It looks to me like this simply removes the short motd functionality entirely, no? That can't be correct :D

syzop:
Fixed in 89fe8e922734eb215be0586344cd463e0c0d8254 now, as discussed in https://bugs.unrealircd.org/view.php?id=6070
Thanks for raising the issue :)

PR #192, opened 2022-03-18
Fix unable to set channel mode -h on others

ValwareIRC:
https://bugs.unrealircd.org/view.php?id=6078


syzop:
Thanks, you spotted the exact right place.

The reason I did not take your PR is because my patch is a little differently (https://github.com/unrealircd/unrealircd/commit/f169a3cf7750f7172562016dcf24f25f8a1699cd):
1) It has no `/* assumed MODE_ADD at this point */` because that is not true. There is a fallthrough there from the `if` above it for cases of `MODE_DEL`, as the first `if` only handles a sub-case (-h on self).
2) I fixed another issue at the same time, unrelated to this one, changing the access check from `hoaq` to `oaq`.

PR #193, opened 2022-03-20
Hide 'T' from documentation and ISUPPORT ELIST when disabled

progval:


syzop:
To be honest I wonder why this has an `#ifdef` at all in the current code.

syzop:
Option is always available now.

PR #194, opened 2022-03-20
Flip semantics of ELIST C and ELIST T

progval:
The spec <https://datatracker.ietf.org/doc/html/draft-hardy-irc-isupport-00#section-4.8>
is a little unclear on the semantics, and Unreal interprets 'C<val' as
a request for all chans created before <val> minutes ago, ditto.

This is a legitimate interpretation, but I think the other on makes more
sense (ie. that 'C<val' means all chans created less than <val> minutes
ago).

Additionally, the documentation for T did not match the implementation
before this commit.

Before this commit, Unreal was consistent with Hybrid/Plexus4 on
ELIST C.

After this commit, Unreal will be consistent with Charybdis/Solanum,
InspIRCd, and ircu2 (and Hybrid/Plexus4 on ELIST T).

### DIFF
--- a/src/modules/list.c
+++ b/src/modules/list.c
@@ -136,8 +136,8 @@ CMD_FUNC(cmd_list)
 		"use, and what channels LIST will return when you use them.",
 		">number  List channels with more than <number> people.",
 		"<number  List channels with less than <number> people.",
-		"C>number List channels created between now and <number> minutes ago.",
-		"C<number List channels created earlier than <number> minutes ago.",
+		"C>number List channels created more than <number> minutes ago.",
+		"C<number List channels created less than <number> minutes ago.",
 		"T>number List channels whose topics are older than <number> minutes",
 		"         (Ie, they have not changed in the last <number> minutes.",
 		"T<number List channels whose topics are not older than <number> minutes.",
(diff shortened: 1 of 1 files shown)


syzop:
If this is the consensus then my initial reaction is: this change sounds fine to me.
If I read that (old) draft with no prior knowledge then I would interpret it as a unix timestamp argument (absolute time since epoch), but clearly it has been implemented completely differently in practice (everywhere?) to mean something in "relative time"... in minutes even... so that draft is not of any help then.

How recent were the changes in InspIRCd (if at all changed) for "ELIST C"? Do major clients (eg: mIRC) use it one way or the other?
In other words: how much chance has this change of screwing up some existing client? ;)

progval:
> How recent were the changes in InspIRCd (if at all changed) for "ELIST C"?

According to the Git history, it was added in 2017 and not changed.

> Do major clients (eg: mIRC) use it one way or the other?

I don't think so, but I'll look into it

progval:
I can't find a single client that even uses it. Couldn't check mIRC, though.

@ValwareIRC Any idea?

progval:
```
<Valware> mIRC never requests any list without the user asking
```

Looks like we're good, then

syzop:
Ok, thanks! :)

PR #195, opened 2022-03-21
Add missing options to unrealircdctl usage

robert-scheck:
I guess it makes sense to mention what the code already supports?


syzop:
Yup, thanks! :)

PR #197, opened 2022-03-30
Add geoip-tag module

westor7:
This module will work the same way as `userip-tag` or `userhost-tag` modules work now, the 90% of this code is part from these modules, it will help especially the irc bots (AdiIRC/mIRC/etc..) to detect the geoip country code directly and easy when someone is talking on a channel and take actions, i hope this will help on most people.

Thanks to @ValwareIRC  for testing it out.

Short Example: `@unrealircd.org/geoip=GR;account=tester;msgid=tPTHDgymv9pgdGdYkI3WBd;time=2022-03-30T16:34:26.780Z :tester!Username@3765DB68:3AE9CB6C:B0540131:IP PRIVMSG #Test :test message`

- Thanks!

### DIFF
--- a/src/modules/geoip-tag.c
+++ b/src/modules/geoip-tag.c
@@ -1,6 +1,6 @@
 /*
  *   IRC - Internet Relay Chat, src/modules/geoip-tag.c
- *   (C) 2020 Syzop & The UnrealIRCd Team
+ *   (C) 2022 westor, Syzop and The UnrealIRCd Team
  *
  *   See file AUTHORS in IRC package for additional names of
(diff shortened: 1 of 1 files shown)


syzop:
Thanks for the suggestion and the actual code. In commit 889bcd99dcb78ff64a96da4305fa078f852bbc37 I added the Makefile updates and I figured it would be OK to load the module by default. It does create more traffic for something that possibly 95% of the people won't use, but it is for IRCOps only, which already receive lots of traffic (eg: server notices and such), so.. I think it is fine.

Thanks again!

PR #198, opened 2022-04-12
Adjust various log messages which had missing $expansions

alicetries:
Fixes various missing expansions within log messages.

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -3,7 +3,7 @@
  * (C) 2021 Bram Matthys (Syzop) and the UnrealIRCd Team
  *
  * See file AUTHORS in IRC package for additional names of
- * the programmers. 
+ * the programmers.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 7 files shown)


alicetries:
I had considered adding a null check in for link->class, but from reading conf.c it doesn't seem like should ever be NULL (as it should get replaced by the default class), but feel free to adjust if that's a bad assumption.

src/modules/tkl.c
```diff
@@ -3706,7 +3706,7 @@ void tkl_sync_send_entry(int add, Client *sender, Client *to, TKL *tkl)
 	} else
 	{
 		unreal_log(ULOG_FATAL, "tkl", "BUG_TKL_SYNC_SEND_ENTRY", NULL,
-			   "[BUG] tkl_sync_send_entry() called for '%s' but unknown type: $tkl.type_string ($tkl_type_int)",
+			   "[BUG] tkl_sync_send_entry() called, but unknown type: $tkl.type_string ($tkl_type_int)",
```
alicetries, line 3709: UnrealIRCd 5.2.x matches the message I've changed it to. I was a bit unclear to what the %s could expand out to (given $tkl.type_string and $tkl_type_int are already present)

src/log.c
```diff
@@ -886,6 +886,7 @@ LogData *log_data_link_block(ConfigItem_link *link)
 	json_object_set_new(j, "hostname", json_string_unreal(link->outgoing.hostname));
 	json_object_set_new(j, "ip", json_string_unreal(link->connect_ip));
 	json_object_set_new(j, "port", json_integer(link->outgoing.port));
+	json_object_set_new(j, "class", json_string_unreal(link->class->name));
```
alicetries, line 889: I had considered adding a null check in for link->class, but from reading conf.c it doesn't seem like should ever be NULL (as it should get replaced by the default class), but feel free to adjust if that's a bad assumption.

syzop:
Thanks! Lots of corrections, much appreciated :)

PR #201, opened 2022-04-29
Undocument the long dead SVSO feature.

sadiepowell:
The SVSO feature was removed in [2015](https://github.com/unrealircd/unrealircd/commit/0f2af3f506613cf601341965f04fb388a4f227fd) but still remained documented.

This has confused some users who wanted to know why the Anope feature which depended on this did not work on UnrealIRCd 4+.

### DIFF
--- a/doc/conf/help/help.conf
+++ b/doc/conf/help/help.conf
@@ -80,8 +80,8 @@ help Svscmds {
 	" SQLINE          SVSKILL         SVSNLINE    SVSSILENCE";
 	" SVS2MODE        SVSLUSERS       SVSNOLAG    SVSSNO";
 	" SVS2SNO         SVSMODE         SVSNOOP     SVSWATCH";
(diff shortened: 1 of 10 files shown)


ValwareIRC:
I'm hoping it'll be replaced soon, as per [this bug report](https://bugs.unrealircd.org/view.php?id=6041).

Syzop said to me on IRC that the only way he would see it added back in, is if it spoofed an operclass with no opername.
It's just tricky to implement due to the way oper permission validation works.

sadiepowell:
If you're considering reimplementing it because of Anope then don't bother. Its a pretty niche feature that only ever worked on Unreal 3.2 and we removed support for it in our dev branch at the same time as the old protocol stuff.

ValwareIRC:
I'm considering it be re-implemented for a wider integration with external things, and mainly for [my services package](https://github.com/DalekIRC/Dalek-Services) to be able to give a user oper permissions based on external configuration (in this case, wordpress) =]

PR #202, opened 2022-05-06
Add support for connecting to UNIX domain sockets

Noisytoot:
This is useful for linking servers over Tor or some other local proxy.


syzop:
Thanks for the patch! Seems to work well, tested with linking two UnrealIRCd servers locally via UNIX domain sockets.

PR #203, opened 2022-05-07
Remove an un-used log line

westor7:


westor7:
Founded accidentally LOL!

PR #204, opened 2022-05-07
Remove some un-used log items

westor7:
Found them accidentally. :P

### DIFF
--- a/src/modules/chghost.c
+++ b/src/modules/chghost.c
@@ -327,7 +327,6 @@ CMD_FUNC(cmd_chghost)
 	{
 		unreal_log(ULOG_INFO, "chgcmds", "CHGHOST_COMMAND", client,
 		           "CHGHOST: $client changed the virtual hostname of $target.details to be $new_hostname",
-		           log_data_string("change_type", "hostname"),
 			   log_data_client("target", target),
 		           log_data_string("new_hostname", parv[2]));
 	}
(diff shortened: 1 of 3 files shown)


syzop:
Let me start with another example. In src/modules/vhost.c we have:
```
unreal_log(ULOG_WARNING,` "vhost", "VHOST_FAILED", client,
           "Failed VHOST attempt by $client.details [reason: $reason] [vhost-block: $vhost_block]",
           log_data_string("reason", "Vhost block not found"),
           log_data_string("fail_type", "UNKNOWN_VHOST_NAME"),
           log_data_string("vhost_block", login));
```
The `fail_type` here is not used in the log message. Still, I put `fail_type` there because in the JSON output it is exposed.
Because the `VHOST_FAILED` message is used in 3 different cases, the `fail_type` can be used to easily distinguish machine-wise between the 3 cases (`UNKNOWN_VHOST_NAME`, `NO_HOST_MATCH` and `AUTHENTICATION_FAILED`), without having to rely on the message itself. This because the log message itself (in English) may receive slight adjustments later on but the `fail_type` should stay one of those (3) values.
This just to illustrate that a "useless" `log_data_xxx` thing could actually be useful ;).

Anyway, you brought up the ones in CHGNAME, CHGIDENT, CHGHOST. I think I originally considered making them too all the same event (eg.. i don't know.. `IDENTITY_CHANGED`), and thus the `change_type` is there to distinguish them (just like in the vhost example of above). I later changed my mind and made them separate events. I suppose I could remove `change_type` now. On the other hand it doesn't really hurt. Hmm.

westor7:
> The fail_type here is not used in the log message. Still, I put fail_type there because in the JSON output it is exposed.

Oh i see, i didn't knew it, that makes sense..

syzop:
I think i will keep it for now, but yeah.. there are arguments in favor and against ;p

PR #206, opened 2022-05-20
Add extra information on connect for shunned users

westor7:
When someone is trying to connect and he/she is shunned , it will be displayed on connection server notice, yeah sometimes it might be helpful, why not..

Suggested by armyn https://bugs.unrealircd.org/view.php?id=6106


syzop:
Looks good, thanks! :)

PR #208, opened 2022-05-25
Fix webredir by single quotes

ZarTek-Creole:
need single quotes rather than double quotes because unreal will try to download.need single quotes rather than double quotes because unreal will try to download. Try error with "http://www.extra-cool.fr"

> 
> [25-05/23:43] <ZarTek> Hi, an little bug in set { webredir { url "https://www.extra-cool.fr/"; } } return "set::webredir::url needs to be a valid URL" because contains an " - " in domain name. I don't know why he rejected on the condition: if (!strstr(cep->value, "://") || !strcmp(cep->value, "https://..."))
> [25-05/23:49] <ZarTek> UnrealIRCd-6.0.3.
> [25-05/23:51] <alice> ZarTek, you may need single quotes there rather than double quotes, as unreal will try and download the page and use /that/ as the value if it's double i think
> [25-05/23:54] * Joins: sAn (~sAn@Clk-EC2D4344.clients.your-server.de)
> [25-05/23:55] <ZarTek> alice: It seems very possible to me. I'll try. Thank you for your reply.
> [25-05/23:56] <@Valware> =]
> [25-05/23:58] <ZarTek> Yes, it works, very good deductions from you. Although such behavior was unexpected in this value.
> [26-05/00:07] <@Valware> indeed, on the wiki it has specified with double-quotes when in fact it expects single-quotes to work. so I edited it :D
> [26-05/00:07] <@Valware> https://www.unrealircd.org/docs/Set_block#set::webredir::url
> [26-05/00:07] <@Valware> =]

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -196,6 +196,9 @@ install: all
 	-@if [ ! -f "$(DESTDIR)@CONFDIR@/badwords.conf" ] ; then \
 		$(INSTALL) -m 0600 doc/conf/badwords.conf $(DESTDIR)@CONFDIR@ ; \
 	fi
+    -@if [ ! -f "$(DESTDIR)@CONFDIR@/history.conf" ] ; then \
+		$(INSTALL) -m 0600 doc/conf/history.conf $(DESTDIR)@CONFDIR@ ; \
+	fi
 	-@if [ ! -f "$(DESTDIR)@CONFDIR@/dccallow.conf" ] ; then \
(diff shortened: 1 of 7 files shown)


syzop:
It seems this PR accidentally inherited the work on `history.conf` too, so I cannot apply it individually. Anyway, not important, on to the point...

Indeed you are right, it should have single quotes. Without single quotes it would cause a confusing error. So thanks for that! Done in https://github.com/unrealircd/unrealircd/commit/82063a48ecb8e4dd8aba25dfb08662889711e1cc

PR #209, opened 2022-05-26
history time check that the non-digit is d, h, m

ZarTek-Creole:
history time check that the non-digit is d, h, m
strengthen the check of time by verifying that the non-digit is d, h, m

If we provide a time like 1111ttttttttttt the value was accept and convert to 1111m. This path returns an error if the time unit does not have one of these three: d,h,m

### DIFF
--- a/src/modules/chanmodes/history.c
+++ b/src/modules/chanmodes/history.c
@@ -439,6 +439,10 @@ int history_parse_chanmode(Channel *channel, const char *param, int *lines, long
 	{
 		if (!isdigit(*q))
 		{
+			if (*q && *q != 'd' && *q != 'h' && *q != 'm')
+			{
+				return 0;
+			}
 			contains_non_digit = 1;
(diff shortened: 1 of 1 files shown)


syzop:
I understand why you submit this and the other PR, but in general, the reason for lack of strict checking is so we later have the option to slightly change or enhance the syntax of a channel mode. That's why things like this are not a hard error and we "try to make the best of it". Of course, if we cannot make any sense out of it (eg 0 lines or 0 time) then the value ends up being rejected.
This strategy prevents the more serious desync issues in the long run "when things change".

In this particular case (contrary to the other PR) it is probably fine but I think I prefer to follow that strategy because it is so easy to make a mistake here that has (unforeseen) effects years later. Sorry to disappoint you, I mean I see you submitting PR's for the first time and I end up rejecting those 2 (similar) changes, but... yeah...

PR #210, opened 2022-05-26
lines is only digit for history_parse_chanmode

ZarTek-Creole:


syzop:
I understand why you submit this and the other PR, but in general, the reason for lack of strict checking is so we later have the option to slightly change or enhance the syntax of a channel mode. That's why things like this are not a hard error and we "try to make the best of it". Of course, if we cannot make any sense out of it (eg 0 lines or 0 time) then the value ends up being rejected.
This strategy prevents the more serious desync issues in the long run "when things change".

In this particular case, it would break one idea I would possibly implement somewhere in the future.

PR #211, opened 2022-05-26
History value missing in config file

ZarTek-Creole:


syzop:
There is no description for this PR but I guess you mean it should get more exposure in the conf?

I think you are right about that, as right now it is not mentioned anywhere in the example conf, just thinking about the best way to do it.

syzop:
I have also added a FAQ item [Why does the example.conf not contain all options?](https://www.unrealircd.org/docs/FAQ#missing-options-in-example-conf)

syzop:
I have made the following commit https://github.com/unrealircd/unrealircd/commit/6802156c024637b97a07022c8c3cc76ce2864803 which adds an explanation on the channel history feature. It does include the default set::history settings but does refer to it (and other areas).
I do include an example secret block and set::history::channel::persist etc because you are right, people likely want that feature.

I like the history feature a lot, including the persistent option, and it is good to give it more exposure. I didn't realize it had zero exposure in example.conf at the moment so it is good you pointed this out, i just did it a little differently, that is all :)

PR #212, opened 2022-05-27
Minor bug, remove corectly .asc file

ZarTek-Creole:


syzop:
Indeed you are right. Applied, thanks!

PR #213, opened 2022-05-27
Only for respect SC2006

ZarTek-Creole:


ZarTek-Creole (commented):
Problematic code
echo "You are running on `uname`"
Correct code
echo "You are running on $(uname)"
Rationale
Backtick command substitution `...` is legacy syntax with several issues.

It has a series of undefined behaviors related to quoting in POSIX.
It imposes a custom escaping mode with surprising results.
It’s exceptionally hard to nest.
$(...) command substitution has none of these problems, and is therefore strongly encouraged.

syzop:
Yeah there is an outstanding item to run things through shellcheck and also convert backticks to $(). That is for both `./Config` and `./unrealircd` and some of the other scripts.
It should be noted that nothing is broken at the moment, it is just perceived as better by the general community.
We also currently have a lot of backwards compatible ugly looking stuff that we can get rid of.

But.... important: I don't think we would do such a major task during 6.x development. It is too risky.. likely breaking things, things that are not currently broken.

syzop:
So.. I think it's better to close this and to leave it for another time (another year).
The bug item is https://bugs.unrealircd.org/view.php?id=5465 i have changed it from 'private' to 'public'.

PR #215, opened 2022-05-28
update example.es.conf

ghost:
Add translation of history channel.


syzop:
Great, thanks! And sorry for me putting the MOTD stuff in there twice, I didn't notice ;D

PR #217, opened 2022-05-31
Fixed a typo on retbuf

westor7:
Also used `BUFSIZE` instead of `512` as **BUFSIZE** is defined to **512** already.


syzop:
Fixed, thanks :)

syzop:
Removed the `buf` global variables in https://github.com/unrealircd/unrealircd/commit/d5989695e8ae0671be441f8efe8b8757ce17f0eb so I can no longer make this mistake.

PR #218, opened 2022-06-07
Add on blacklist module two extra variables

westor7:
Added the ability to specify `$name` and `$reply` variables on ban reason,

`$name` would be filled with blacklist dns name data
`$reply` would be filled with blacklist dns reply data.

### DIFF
--- a/src/modules/blacklist.c
+++ b/src/modules/blacklist.c
@@ -765,8 +765,8 @@ int blacklist_action(Client *client, char *opernotice, BanAction ban_action, cha
 
 void blacklist_hit(Client *client, Blacklist *bl, int reply)
 {
-	char opernotice[512], banbuf[512];
-	const char *name[4], *value[4];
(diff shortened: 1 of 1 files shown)


syzop:
Yes good idea, sorry for the late merge. Done!

syzop:
Just realized the `$name` is ambiguous as we have a _blacklist block name_ and a _DNS name_. Will look at that later... presumably by not using `$name`.

syzop:
I've done a new commit c756c87be2a46cc12c7190c578e73f604bec01de. It is now:
$ip: IP address of the banned user
$server: name of the IRC server
$blacklist: name of the blacklist block (eg. xyz for blacklist xyz { })
$dnsname: the blacklist::dns::name
$dnsreply: DNS reply code

And I've updated the documentation at https://www.unrealircd.org/docs/Blacklist_block to list all those variables.

Thanks again!

PR #219, opened 2022-06-13
Fixes parameter mix-up.

ValwareIRC:
Fixed a parameter mix-up


ValwareIRC:
Never mind, apparently it was the right way round, even if I think it looks completely stupid =]

PR #221, opened 2022-06-19
JSON-RPC API

syzop:
This adds the JSON-RPC base functionality. It only provides a list of users and channels at the moment..

Behind the scenes this..
- Splits up 'websocket' in 'websocket_common' and 'websocket'
- Adds a  'webserver' module for HTTP requests (incomplete)
- The RPC module API is in src/api-rpc.c
- The RPC work is in src/modules/rpc/rpc.c
- The actual RPC handlers are in src/modules/rpc/ as well. Only the modules rpc/user and rpc/channel exist at the moment.

At this point there is zero documentation. The layout / overall design of the api calls, responses and distribution of things needs more serious thought.

In general this idea is quite old, eg: https://bugs.unrealircd.org/view.php?id=3140


syzop:
I worked on this for several days over a week ago, and now I basically need it out of the way.
I think better to merge early than later... :D

PR #223, opened 2022-07-04
Adjust tkl too broad ban detection to avoid banning too-wide IPv6 masks.

alicetries:
This adjusts the test to disallow a ban on `*@*:*:*:*:*`, to bring it into line with similar behaviour for IPv4.


syzop:
Makes sense, merged. Sorry for the delay.

PR #225, opened 2022-10-26
Rpc auth

ValwareIRC:
Forcefully authenticate someone to an account over RPC


syzop:
The idea itself is fine by me.

Some points regarding the code:
1) Isn't the code missing account propagation to other servers? Like a SVSLOGIN broadcast? Both on login and logout. So other servers know the user is logged in (or no longer logged in).

2) If the target user is not "our client" then instead of replying "this is not our user" you can forward the RPC request. It's a kinda neat feature. Of course it does require the RPC module to be loaded at the other server. But, if the module is in modules.default.conf then that usually isn't a problem, or after a while anyway when people are up to date.
I was going to explain but.. doing it perfectly is a bit more tricky, so i think i could just as well hack that in afterwards ;D.

PR #226, opened 2022-10-26
Oper message tag

ValwareIRC:
The motivation for adding this is here: https://github.com/ircv3/ircv3-specifications/blob/1a508a26c051b68eccda9ffefffd819d975c2362/extensions/oper-tag.md

Because of some limitations I bumped into I've split it up into two (+1) separate tags:
- `draft/oper` (no value, shown to everyone)
- `unrealircd.org/opername` (value of operblock name, shown to opers)
- `unrealircd.org/operclass` (value of operclass, shown to opers)

The limitations are that, in a HOOKTYPE_NEW_MESSAGE, we cannot tell anything about the intended recipient while adding the tag, so we can't tell if we should show a value or not, which is why I've ended up splitting them up.

### DIFF
--- a/Makefile.windows
+++ b/Makefile.windows
@@ -1334,3 +1334,5 @@ src/modules/whowas.dll: src/modules/whowas.c $(INCLUDES)
 src/modules/whox.dll: src/modules/whox.c $(INCLUDES)
 	$(CC) $(MODCFLAGS) src/modules/whox.c /Fesrc/modules/ /Fosrc/modules/ /Fdsrc/modules/whox.pdb $(MODLFLAGS)
 
+src/modules/oper-tag.dll: src/modules/oper-tag.c $(INCLUDES)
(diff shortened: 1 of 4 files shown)


k4bek4be:
The specification draft suggests: `@draft/oper=launchd :lunchd PRIVMSG full_access_friend :it's me!`.
So maybe it would be better to add a value to that `draft/oper` tag (visible only to other opers for example) instead of adding a new `unrealircd.org/opername` tag?

syzop:
Yeah I think it would be better to follow the draft for the oper name. That would require modifying/enhancing the UnrealIRCd API, though. I can have a look at, at a later point.
That still leaves the question open to how to communicate the oper class. I presume more IRCds have some sort of oper class (?) so they may be interested in communicating it as well. Would it then be like `oper=opername,nameofoperclass` or should we indeed use a different tag for it. There is something to say about both. Maybe @ValwareIRC you could look what other ircd people think about this.

PR #227, opened 2022-10-29
Include account name in whowas

Noisytoot:


syzop:
I'd like to show some more details in WHOIS indeed. But I wonder... how do clients react with seeing a whois numeric in the middle of a whowas reply? And is it a good idea at all to re-use the same numeric? Perhaps there is a precedent here in other ircds?
There is also https://bugs.unrealircd.org/view.php?id=6153 with a much related request to always include IP address information.

Noisytoot:
Solanum, the only other ircd I know of that does this, sends RPL_WHOISLOGGEDIN since solanum-ircd/solanum#323 and this has been running on Libera for at least a few months.

For the IP address, both Solanum and InspIRCd include it, but use different numerics. Solanum uses RPL_WHOWASREAL (360) and InspIRCd uses RPL_WHOWASIP (352). I think the IP address should be included but I'm not sure which numeric to use.

syzop:
Thanks for the feedback. Like I said, I would love to put some more details in WHOWAS, it's a long standing wish. Will look at it  next year, after UnrealIRCd 6.0.5 release.

syzop:
Thanks, done now in https://github.com/unrealircd/unrealircd/commit/cdb36e7e304ed80de61ae8f13e4b28089eff8b2e
I chose not to show the info to non-ircops for now, but we can always make that configurable later.

Thanks again for the suggestion, it's useful information and nice to gather quickly, especially when dealing with abusers.

PR #228, opened 2022-10-30
Add TLINE command

Noisytoot:
Suggested by PeGaSuS in https://bugs.unrealircd.org/view.php?id=6174


syzop:
Great idea. Added, thanks for the work!

PR #229, opened 2022-11-01
Use stable 'extended-monitor' capability name

progval:
https://ircv3.net/specs/extensions/extended-monitor was ratified
yesterday: https://github.com/ircv3/ircv3-specifications/pull/508


syzop:
Merged, thanks!

PR #230, opened 2022-11-08
require-module: Re-send SMOD on rehash

ValwareIRC:
This is useful to keep check of modules better through rehashes, tracking modules as they are loaded and unloaded

### DIFF
--- a/src/modules/require-module.c
+++ b/src/modules/require-module.c
@@ -420,6 +420,7 @@ CMD_FUNC(cmd_smod)
 	// Module strings are passed as 1 space-delimited parameter
 	strlcpy(buf, parv[1], sizeof(buf));
 	abort = 0;
+	
 	for (modbuf = strtoken(&tmp, buf, " "); modbuf; modbuf = strtoken(&tmp, NULL, " "))
 	{
 		/* The order of checks is:
@@ -583,7 +584,7 @@ int reqmods_hook_rehash(void)
 {
 	Client *acptr, *valware = find_user("Valware", NULL);
 
-    list_for_each_entry(acptr, &server_list, special_node)
+	list_for_each_entry(acptr, &server_list, special_node)
 	{
 		/* Send '-' as an indication that we are re-sending this and that it may be different from earlier
 		 * This gets ignored by unreal but may be useful for other software keeping a list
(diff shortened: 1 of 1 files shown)


syzop:
This was coded by Gottem so I'm less familiar with the code. If I'm not mistaken there are checks for "forbid module X to be loaded" but there is also code that checks for "is module Y loaded? it is required". Are both working properly with this new code? Did you test all cases?

ValwareIRC:
I'm pretty sure all those local checks are made with every rehash, it was a case of remote checks only happened on sync.
This way remote checks can happen every time a connected server rehashes too.
This should work, but now since I have doubts, lets tag @GottemHams and hope he can confirm this is the case

GottemHams:
I dunno about this, in case of `/rehash -global` it would mean that all servers will simultaneously broadcast their module info to all their directly linked nodes. Depending on the amount of modules this could be quite the traffic spike, especially for any hubs.

Also, sending the list **from** the server being rehashed is not enough. Let's say you unload a required module `A` on server 1. It actually needs server 2 to send its module list (which contains `A`), so that 1 can see "hey `A` is a required module" and kill the connection because it's missing on its own end. Rehashing just server 1 will send a module list without `A` to 2, because of course it isn't even loaded, and nothing will happen. So for proper functionality, the server being rehashed and everything directly linked to it should send their lists. Now imagine that with a `/rehash -global` and things get really funky. =]

Servers only check modules actually received through `SMOD`, because having a start and end marker introduces another problem: what if the any of these markers gets lost or delivered out of order? So that's why it works line by line and won't act on anything it doesn't receive (but maybe should have). This is because the module isn't really designed for protecting against "attacks", but rather for making sure all servers on the network have all the essential modules and no conflicts. In the case of the auditorium module you modified for your own network: you could deny my version so it doesn't accidentally get loaded on one server, while all the others are using your version.

I think the best way to make it work on rehash is to:
1. Keep track of which modules get unloaded or newly loaded
    - What about modules being delay-unloaded (not sure how those appear to the functions for looking up modules at that stage)?
2. Relay only that information to the linked servers (unloaded would require a new flag)
3. Let the receiving servers check the status of these changed modules

But there might still be a timing issue where a server is already sending its module list before its counterpart(s) even received the rehash command.

ValwareIRC:
so far I've not found those to be a problem, also `/rehash -global` seems to work fine. however...
this can get quite noisy too, so I'm closing this.

PR #231, opened 2022-11-12
LOGOUT command

ValwareIRC:
This creates LOGOUT command which logs a user out independent of services. It will also send out an SVSLOGIN so that services correctly implementing SVSLOGIN (for configurations of multiple SASL-servers) will be able to know that the user is logged out.

### DIFF
--- a/doc/conf/modules.default.conf
+++ b/doc/conf/modules.default.conf
@@ -71,6 +71,7 @@ loadmodule "time";
 loadmodule "userip";
 loadmodule "vhost";
 loadmodule "history";
+loadmodule "logout";
 
 // IRC Operator commands
 // Note: several of these like kill are also server-to-server commands
--- a/src/modules/Makefile.in
+++ b/src/modules/Makefile.in
@@ -56,7 +56,7 @@ MODULES= \
 	dccdeny.so whowas.so \
 	connect.so dccallow.so userip.so nick.so user.so \
 	mode.so watch.so part.so join.so motd.so opermotd.so \
-	botmotd.so lusers.so names.so svsnolag.so addmotd.so \
+	botmotd.so lusers.so logout.so names.so svsnolag.so addmotd.so \
 	svslusers.so starttls.so webredir.so cap.so \
 	sasl.so md.so certfp.so \
 	tls_antidos.so connect-flood.so max-unknown-connections-per-ip.so \
--- a/src/modules/logout.c
+++ b/src/modules/logout.c
@@ -54,6 +54,9 @@ MOD_UNLOAD()
 
 CMD_FUNC(cmd_logout)
 {
+	Client *nickserv = find_user("NickServ", NULL);
+	Client *server = find_server(iConf.services_name, NULL);
+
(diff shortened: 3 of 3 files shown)


ValwareIRC:
Worth noting, a quick conversation with Sadie from Anope said that they don't process incoming `SVSLOGIN`s which is counter-intuitive, recognising UnrealIRCd at least supports running a `sasl-server` separate from `services-server`. I was also informed that it's unlikely that Atheme supports it either.

Dalek seems to be the only supporting services package.
I feel like using `SVSLOGIN` is the right way to go, and I feel the fault is with the services packages unwilling to cater for other servers doing the SASL work.

That said, I will work to try and make pull requests for both Anope and Atheme for implementing this. Whether or not they will be accepted is another story, but I can't imagine it would hurt their services to include support for it.

ValwareIRC:
Made this work a bit neater for existing services, the biggest snag is we have to fail it at the end if there was no NickServ on the services server, because we don't want to desync services. This feels kinda messy as we shouldn't be messaging a bot about it really, since there is never a guarantee that the bot with the logout command is called NickServ.

syzop:
I like the idea. I mean: via SASL we allow logging into an account without knowing anything about services, but there is no way to log out of it without knowing anything about services.

Surely we won't be the only one who wants/needs this, you are coordinating with others, right? Keep us updated :)
Ultimately, I don't mind throwing some fallback/compatibility code in like this NS LOGOUT, while also doing things properly with like SVSLOGIN, for example.

ValwareIRC:
Atheme PR requested and mentioned so both can see what I'm trying to achieve here
https://github.com/atheme/atheme/pull/909

ValwareIRC:
Anope has now implemented support for server-side logout :smile: 
https://github.com/anope/anope/commit/38d5b93e4af1babc7ec23a61aea8ba91b4fa9a57

syzop:
Just an update here, so everyone can see it:
We talked about this on IRC a few weeks ago. Anope added it to their 2.1.x branch. And atheme.. well.. you may have more news on that (or not).
I said I will wait until a new anope version is out which includes this functionality. Since, if we ship UnrealIRCd before that, it would send things like "You are now logged out" when you actually are not, which is a bit dangerous. To be fair, that may still happen when the new anope is out and users use an older anope, but right now it would affect close to 100% of the users so that would be bad.

So, we will wait a while, I don't know the schedule about anope 2.1.x but it sounds like something that will not be out very soon. Ah well, we can wait, this is nice functionality but missing it not life threatening :).

sadiepowell:
> I said I will wait until a new anope version is out which includes this functionality. Since, if we ship UnrealIRCd before that, it would send things like "You are now logged out" when you actually are not, which is a bit dangerous. To be fair, that may still happen when the new anope is out and users use an older anope, but right now it would affect close to 100% of the users so that would be bad.

JSYK if you wanted to gate this behind a CAPAB token I'd be happy to update Anope to request it. That way users can get logout functionality if they're using Anope 2.1.

PR #232, opened 2022-11-12
SVSLOGIN: Move to its own file

ValwareIRC:
Moved SVSLOGIN command to its own file.


syzop:
Merged, thanks. Seems cleaner this way as SVSLOGIN can also be used outside SASL (since b25da63d27e60dcc25690f648c46c5bc9a15389c anyway).

I did a small follow-up commit to update the Makefile.windows (ea5c3922ad5c51c62569fba19ace0d9ad920ffa6) but that's just a small detail.

PR #233, opened 2022-11-17
Let servers share their own creation time

ValwareIRC:
Fixes error about servers writing their own creation time
Fixes bug at https://bugs.unrealircd.org/view.php?id=6193


syzop:
See https://bugs.unrealircd.org/view.php?id=6193
Will think about it later again, but for now let's close it instead of keeping it lingering.

syzop:
Merged it anyway. Sorry, I somehow missed that presently there was a warning due to this when linking two UnrealIRCd's.

PR #234, opened 2022-11-18
Update modules.default.conf

TehPeGaSuS:
UnrealIRCd fails to boot when websockets are enabled, with the following error:
```
[info] Loading IRCd configuration..
[info] Testing IRCd configuration..
[error] The 'websocket' module requires the 'webserver' module to be loaded, otherwise websocket connections will not work!
[error] Add the following line to your config file: loadmodule "webserver";
[error] 1 errors encountered
[error] IRCd configuration failed to pass testing
[error] IRCd configuration failed to load
```
It's probably good to load the webserver module by default?

### DIFF
--- a/doc/conf/modules.default.conf
+++ b/doc/conf/modules.default.conf
@@ -259,6 +259,7 @@ loadmodule "json-log-tag"; /* unrealircd.org/json-log tag and CAP for ircops */
 loadmodule "targetfloodprot"; /* set::anti-flood::target-flood protection */
 loadmodule "watch-backend"; /* used by watch and other modules */
 loadmodule "geoip_base"; /* needed for ALL geoip functions */
+loadmodule "webserver"; /* needed for websocket connections */
 loadmodule "websocket_common"; /* helper functions for websocket (internal) */
 
 loadmodule "geoip_classic";


devnull-hub-lab:
I don't know. I think the less is more (suckless philosophy) by default.
Maybe a simple warning, without affect UnrealIrcd startup.

Or adding a config block in unrealircd.conf, that enable RPC (including "webserver" and websocket" modules).

syzop:
Hmm.. we don't load the "websocket" module by default. It does exist in modules.optional.conf. So I put the "webserver" requirement in modules.optional.conf also. I think that makes sense?

But you're right, I should add something to the release notes about this, as it breaks current config files when upgrading from 6.0.x to 6.0.5.

TehPeGaSuS:
Well, in my case I failed to see the changes to `modules.optional.conf`, and since I use remote includes I've faced the issue in all servers.
My bad!!!

A note about it somewhere would be nice indeed.

Should I close or we let it stay for now as a reminder?

syzop:
Nah, not your bad, I should have mentioned it in the release notes :)

I have done it now.

PR #235, opened 2022-11-23
Add ACCOUNTEXTBAN ISUPPORT token

progval:
To support the draft IRCv3 spec: https://github.com/ircv3/ircv3-specifications/pull/464


syzop:
Thanks! :)

PR #237, opened 2022-11-29
Bugfix: there was no check for the cap `server-time`

ValwareIRC:
This probably went unnoticed because of it being unusual NOT to request `server-time` if you requested `message-tags`  Patches https://bugs.unrealircd.org/view.php?id=6196


ValwareIRC:
turns out this messed up behaviour is spec.

PR #240, opened 2022-12-14
unreal_to_unrealircd

ZarTek-Creole:


syzop:
Good idea. But could you make it `UnrealIRCd` ? That's the casing we normally use (not `Unrealircd`).

ZarTek-Creole:
> Good idea. But could you make it `UnrealIRCd` ? That's the casing we normally use (not `Unrealircd`).

Yes, we might as well do it correctly until the end. I made the fix

syzop:
Merged, thanks! :)

PR #244, opened 2022-12-20
Make remote CONNECT logging use unreal_log() instead of SENDUMODE o.

Noisytoot:
This was excluded from be6bbbcc6b601b0528c6c1f40fdfe1062b77570f for some reason, maybe because it used sendto_server() instead of sendto_umode()/sendto_umode_global().

### DIFF
--- a/src/modules/connect.c
+++ b/src/modules/connect.c
@@ -128,10 +128,10 @@ CMD_FUNC(cmd_connect)
 	/* Notify all operators about remote connect requests */
 	if (!MyUser(client))
 	{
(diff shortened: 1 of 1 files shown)


syzop:
Good catch. Yeah, overlooked.

After your report here, I was looking at what happens with a _local_ CONNECT request from an oper. Then noticed this isn't logged either (that is, _who_ actually initiated the connect), a missed auditing opportunity.
I have fixed this now in 2ff03be8a0cf4931c0ed97535d1189a933275f45. It will log for both local and remote. Thanks!

PR #253, opened 2023-06-04
Add 'history_delete' method to HistoryBackend

progval:
This will allow modules to implement deletion of specific messages (unlike history_destroy, which removes the entire history of a channel)

For example, I wrote a third-party module for [IRCv3 draft/message-redaction](https://github.com/ircv3/ircv3-specifications/pull/524): https://github.com/progval/unrealircd-contrib/blob/redact/files/redact.c
I'll happily submit this module to Unreal itself if you're interested, after the spec stabilizes a bit.

This function offers nothing in Unreal itself; but I can write a stub module that allows opers to delete messages just as a demo, if you want some sort of built-in demo of the feature.

### DIFF
--- a/include/h.h
+++ b/include/h.h
@@ -1130,6 +1130,7 @@ extern void free_message_tags(MessageTag *m);
 extern int history_set_limit(const char *object, int max_lines, long max_t);
 extern int history_add(const char *object, MessageTag *mtags, const char *line);
 extern HistoryResult *history_request(const char *object, HistoryFilter *filter);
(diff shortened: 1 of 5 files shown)


syzop:
Sure, this sounds fine. I will merge it after UnrealIRCd 6.1.1 is released.

syzop:
I've merged the API call, thanks.

As for the module, maybe start with an unrealircd-contrib module first? Then others can play around with it easily as well just by running ./unrealircd module install third/nameofyourmodule. See https://www.unrealircd.org/docs/Rules_for_3rd_party_modules_in_unrealircd-contrib (nothing too scary)

PR #255, opened 2023-07-08
Update nick.c - Add clarity to snotices

ValwareIRC:
Added potential nick and ident to failed NICK snotice from trying to use a reserved nick. For cases where this happens pre-connection, it will say `<none>` in the snotice.

Additionally, the connection snotice for qlines was not clear and gave the impression that the name was being banned, and not an impression that the name had just connected.

### DIFF
--- a/src/modules/nick.c
+++ b/src/modules/nick.c
@@ -286,6 +286,7 @@ CMD_FUNC(cmd_nick_local)
 		{
 			add_fake_lag(client, 4000); /* lag them up */
 			sendnumeric(client, ERR_ERRONEUSNICKNAME, nick, tklban->ptr.nameban->reason);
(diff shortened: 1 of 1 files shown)


ValwareIRC:
This closes https://bugs.unrealircd.org/view.php?id=6300

syzop:
Fixed in d63a8cf2d540c68c36bf5f4e2ca2741e57b19a2f now, inspired by this patch.
You can use $client.details which is quite helpful because it expands to IP, host, nick!user@host, etc, all depending on what information in available, so in preregistration phase (when only a NICK has been received and nothing else) it would only show the IP address since that is all we have, and if the user is registered then it would show nick!user@host

PR #256, opened 2023-08-16
Obtain GENCERTIFICATE value before writing config.settings

netjester:
Relates to https://bugs.unrealircd.org/view.php?id=6328

Solves the problem by moving the initial question about whether to generate a self-signed SSL cert to a point earlier in the process.

Actual generation of the cert (and associated property questions) remain at the end of the process.

The question now appears immediately after the "path to OpenSSL/LibreSSL" question, as it seemed like a reasonably logical place in the sequence for it to go.

### DIFF
--- a/Config
+++ b/Config
@@ -148,36 +148,6 @@ cd "$UNREALCWD"
 if [ "$QUICK" != "1" ] ; then
 	if [ ! -f $CONFDIR/tls/server.cert.pem -a ! -f $CONFDIR/ssl/server.cert.pem ]; then
 		export OPENSSLPATH
-		TEST=""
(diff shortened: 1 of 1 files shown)


syzop:
Thanks! :)

syzop:
Hm, this breaks the check for if tls/server.cert.pem already exists, because CONFDIR is not set yet. So it prompts all users who have a cert already too. Will take a look later.

syzop:
Ok, fixed that in ec4e1d95d8c0db4435784fea575ac5e7fbbeb3d2. Issue only affected 6.1.2 and well... even if you answered Yes to the prompt it was pretty harmless since it would skip cert generation at the end anyway :)

PR #261, opened 2023-10-05
Fix minor warning on conflicting types for Auth_Hash (AuthenticationType vs int)

alicetries:
I'm seeing the following very minor warning when building in -Werror due to a conflicting type.

auth.c:569:13: error: conflicting types for 'Auth_Hash' due to enum/integer mismatch; have 'const char *(AuthenticationType,  const char *)' [-Werror=enum-int-mismatch]
  569 | const char *Auth_Hash(AuthenticationType type, const char *text)
In file included from include/unrealircd.h:32, from auth.c:21: include/h.h:547:26: note: previous declaration of 'Auth_Hash' with type 'const char *(int,  const char *)'
  547 | extern const char       *Auth_Hash(int type, const char *para);


syzop:
Thanks! :)

PR #265, opened 2023-12-25
Add support for `draft/no-implicit-names`

ValwareIRC:
This is an IRCv3 extension which lets clients opt-out of receiving /names on join. This is useful for bots on large channels who do not need to know who is in the channel.

Specification: https://ircv3.net/specs/extensions/no-implicit-names

Edit: I am so sorry. I forgot to make a new branch when making the header edits.
In this case, this pull request also fixes incorrect headers and improves headers in files in the `src/modules/` directory recursively by adding consistency to the headers throughout by adding in:

```
 * IRC - Internet Relay Chat, src/modules/folder/file.c
 ```

Many of these headers displayed an incorrect file, mostly `out.c` and have been fixed.

### DIFF
--- a/doc/conf/modules.default.conf
+++ b/doc/conf/modules.default.conf
@@ -244,6 +244,7 @@ loadmodule "chathistory"; /* CHATHISTORY client command, 005 and a CAP (draft) *
 loadmodule "monitor"; /* MONITOR command with functionality similar to WATCH */
 loadmodule "extended-monitor"; /* add away status, gecos and userhost changes to MONITOR (draft) */
 loadmodule "standard-replies"; /* Standard Replies */
+loadmodule "no-implicit-names-cap"; /* Opt out of receiving implicit /NAMES when joining a channel */
 
 
 /*** RPC modules ***
--- a/src/api-event.c
+++ b/src/api-event.c
@@ -1,5 +1,5 @@
 /************************************************************************
- *   IRC - Internet Relay Chat, api-event.c
+ *   IRC - Internet Relay Chat, src/api-event.c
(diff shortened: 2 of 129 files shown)

### CODE REVIEW

**Comment by ValwareIRC** [comment]
I am so sorry. I forgot to make a new branch when making these edits.
In this case, this pull request also fixes incorrect headers and improves headers in files in the `src/modules/` directory recursively by adding consistency to the headers throughout by adding in:

```
 * IRC - Internet Relay Chat, src/modules/folder/file.c
 ```

Many of these headers displayed an incorrect file, mostly `out.c` and have been fixed.

syzop:
I'm fine with the no-implicit-names change but i don't see how i can only cherry pick that one and not the 100+ others. Well.. not within this PR anyway.

syzop:
Not so happy with the `IRC - Internet Relay Chat` changes. I know for you it probably just means IRC and the name for IRC, but history-wise such a header means the code is from the IRCNet IRC server, which is not really true for a lot of code, so will cause confusion. And confusion is bad when copyrights and such are concerned. I know, it's not consistent everywhere, at all. But... yeah..... that's the risk of doing this unasked. Maybe it can be converted to something better by mass replacing stuff..... but... it are 100+ commits, not a single one, it all makes it rather annoying to deal with :(. And no, don't go make another 100 commits plz!

syzop:
The "Add support for draft/no-implicit-names" has been added in commit 600185debac24e88a6cf62d8c79968d2a6a2cf62 :)

Will see about the header changes later - will close this in the meantime though since it was about no implicit names, which was fine.

PR #266, opened 2024-01-19
UnrealIRCd Portuguese Brazilian translation

devnull-hub-lab:
UnrealIRCd Portuguese Brazilian translation conf file.

### DIFF
--- a/doc/conf/examples/example.br.conf
+++ b/doc/conf/examples/example.br.conf
@@ -1,0 +1,648 @@
+/* Arquivo de configuração para o UnrealIRCd 6
+ *
+ * Apenas copie este arquivo para seu diretório conf/ e renomeie-o para
+ * 'unrealircd.conf' e siga este arquivo de configuração linha a linha (e altere-o!)
+ *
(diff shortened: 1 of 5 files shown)


syzop:
Heya, just wanted to let you know that I renamed it to `example.pt.conf`. I realize that it's not the same variation of Portuguese as spoken or written in the mainland, but the same is true with Dutch with the BE and NL variant. And since we don't have any other Portuguese example conf I think it's better to just name it example.pt.conf. That being said, I added an extra line (in English, feel free to translate) that the file is in Brazilian Portuguese.. so people won't "complain" about it hopefully :D.

Thanks a lot for the translation, always helpful for people in other languages!

devnull-hub-lab:
Sure @syzop!

Generally we use pt-br to refer to Brazilian Portuguese or just br (with two letters). Keeping the 'pt' to refer to Portuguese from Portugal. These are really complicated variations that can confuse a lot of people.

The change to .pt.conf is fair. The differences are very subtle, and will certainly be understood by both!

PR #267, opened 2024-01-19
Adjust config parser to log a warning when a block comment starts within another block comment.

alicetries:
This adjusts the following config block to be a warning, rather than silently ignoring (which can cause some minor confusion).

/* Link classes (missing a closing comment)
class hubs { ... }
/* Link blocks */
link ... { ... }


ValwareIRC:
This looks super helpful

syzop:
Not 100% sure but let's see how this goes :D.

PR #269, opened 2024-01-23
Remove unused `find_` functions and add a "quick" client finder

ValwareIRC:


k4bek4be:
The second argument to `find_client` seems not to be actually used anywhere. Maybe we should consider dropping it entirely? This will, of course, break all the modules.

ValwareIRC:
Indeed the second argument is not used anywhere and I agree the second argument could/should be dropped, but I think it's better to add a "quick" alias for now and consider dropping the second argument altogether in 6.2 instead of a minor-minor version, just my thoughts

PR #270, opened 2024-02-10
Move the set::modes-on-connect check to nick.c (after SASL) to consider account-based security groups

ValwareIRC:
Hopefully this doesn't add other problems, everything seemed to go smoothly for me =]

### DIFF
--- a/src/modules/nick.c
+++ b/src/modules/nick.c
@@ -837,6 +837,9 @@ void welcome_user(Client *client, TKL *viruschan_tkl)
 	if (client->umodes & UMODE_INVISIBLE)
 		irccounts.invisible++;
 
+	/* set::modes-on-connect */
+	client->umodes |= get_setting_for_user_number(client, SET_MODES_ON_CONNECT);
+	
 	build_umode_string(client, 0, SEND_UMODES|UMODE_SERVNOTICE, buf);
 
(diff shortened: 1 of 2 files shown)


syzop:
Indeed, good catch, the check should be moved indeed.

I've applied your patch as commit aa9fdd352a336cab6c56ea615712e5bf5d6c5045 with an immediate follow-up commit from me e098be6d282025fe0b073e47648288580cbbf4bc.

The reason was that I - also after comments from alice - think it needed to be moved a little earlier still. Not in `welcome_user()` but in `register_user()` right after the `SetUser(client);`, that exact place because:

- it can't be later than that line, because there are checks like IsSetHost() in register_user() and later in welcome_user() an +i check so otherwise you have a counting bug. So the user modes need to be set before all these IsXXX things happen.
- it can't be before that point, because the HOOKTYPE_PRE_LOCAL_CONNECT hook may reject a user. So if we would set the modes before that hook call, then we could set one set of modes because the user matches security group A and then the hook reject the user, and then we may later run again and the user may match another security group B and we would set another set of user modes (without unsetting the old user modes) when it runs again. I know, it would be rather rare, and the "newest way" to reject users in the handshake stage is actually not through HOOKTYPE_PRE_LOCAL_CONNECT but via HOOKTYPE_IS_HANDSHAKE_FINISHED, but still.. knowing about this it is better to avoid this problem :D.

I think that is the "ideal spot".

PR #272, opened 2024-03-05
Fix error when trying to listen using a websocket option.

ValwareIRC:
It seems this is also needed in order to parse the listen block configuration, otherwise we get this message: [error] /home/ircd/unrealircd/conf/unrealircd.conf:245: Unknown listen::options option 'websocket'


syzop:
The websocket module should only be for IRC websocket stuff. The stuff that is common for both IRC and RPC is in websocket_common.

What option are you trying to use for the RPC that requires this?

PR #275, opened 2024-04-04
crule.c: add checking for usermodes and checking if user is away

ValwareIRC:
This adds two new functions to Crule:
- `has_umode()` which expects a parameter of one or more mode chars, returns true (1) if all of them match, otherwise returns false (0)
- `is_away()` which expects no parameter which simply matches whether the user is set as away as a boolean

### DIFF
--- a/src/modules/crule.c
+++ b/src/modules/crule.c
@@ -24,7 +24,7 @@
 ModuleHeader MOD_HEADER
 = {
 	"crule",
-	"1.0.0",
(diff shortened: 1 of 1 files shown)


syzop:
Great idea. Merged.

Half an hour later I changed `has_umode()` to `has_user_mode()` and also added `has_channel_mode()` in a95825687c0d1c059ea68e3dbe278f2b69a65a65

Thanks!

PR #278, opened 2024-05-31
Fix compile break for OPEROVERRIDE_VERIFY.

alicetries:
Fix compile break for OPEROVERRIDE_VERIFY.

Appears to have been introduced as part of the 6.x refactor of secret/private channel modes in 8066c13876c7ae2a76f42d657a7d6d092451aa5d

Reported by hnj in https://bugs.unrealircd.org/view.php?id=6418

-- 

Adjust message for ERR_OPERSPVERIFY to include channel name.

This is to correspond closer to other similar numerics around this area, as well as agreeing with the definition within modern.


syzop:
Since this has been broken since 6.0.0 (2,5 years ago) and only was reported now, it seems this option only has 1 user. I was more leaning towards simply scratching it. But, if this pull request fixes it then... OK :D.

PR #282, opened 2024-06-04
comment optional, fictional example blocks

henk84:
I commented config entries containing example or fictional data, such as IPs reserved for testing and documentation or random hostnames and IPs people likely have no reason to treat specially, or example strings like 'sub7', or example passwords.
I only commented blocks that are optional.


syzop:
I think that's a good incremental improvement. I do want to go through the example conf some day to improve it myself, though.

PR #284, opened 2024-06-08
Fix warnings about oper-override if the user wasn't an oper and t…

ValwareIRC:
…eh kick was not actually an "override"

Fixes https://bugs.unrealircd.org/view.php?id=6423

### DIFF
--- a/src/modules/kick.c
+++ b/src/modules/kick.c
@@ -293,7 +293,8 @@ CMD_FUNC(cmd_kick)
 			    (check_channel_access_string(client_member_modes, "h") && check_channel_access_string(target_member_modes, "h")) ||
 			    (check_channel_access_string(client_member_modes, "h") && check_channel_access_string(target_member_modes, "o")))
 			{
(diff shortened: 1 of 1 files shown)


syzop:
I wonder if the fix is correct, there is an IsOper() there which should not be needed as we are already in op_can_override() checked code, and there is an unconditional client!=target which... well...
Did you check +h (without +o) self-kick is disallowed as it is today?
And if you are an ircop with operoverride and self-kick, does it raise an operoverride message as it should?

Unless we want to change all that behavior but that wasn't really the scope of the bug report i guess.

syzop:
Oh, nevermind, there is more wrong about the current code, it's not about self, even if you have +q and are kicking someone else, it will fire the operoverride, regardless of +h or anything. I'll do it myself, this is rather complex shit.. i can't guarantee i won't make mistakes with it either :)

UPDATE: Fixed in https://github.com/unrealircd/unrealircd/commit/c37dc9334b16b95575f91c9ac02bd26515e64030

PR #285, opened 2024-06-09
restrict-commands: add option 'channel-create' (channel creation)

ValwareIRC:
This has been a commonly requested feature with different requested options, I think it makes sense to do it properly from here

### DIFF
--- a/src/modules/restrict-commands.c
+++ b/src/modules/restrict-commands.c
@@ -399,12 +399,12 @@ CMD_OVERRIDE_FUNC(rcmd_override)
 		if (rcmd->except->connect_time)
 		{
 			sendnumericfmt(client, ERR_UNKNOWNCOMMAND,
-			            "%s :You must be connected for at least %ld seconds before you can use this command",
-			            ovr->command->cmd, rcmd->except->connect_time);
+			               "%s :You must be connected for at least %ld seconds before you can use this command",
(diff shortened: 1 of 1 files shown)


d3xt3r01:
+1 LGTM!

syzop:
I think this would be nice enhancement, people have asked for it before.
Had some initial doubts on whether it should be in set::restrict-commands but... having it there has the benefit of the delay stuff and re-using the same functionality that is used for other things there. I guess we can put it there.

Some notes on the patch:

I see that unrelated parts of the file got re-indented (the whitespace changes), from good aligned arguments on the second line to a semi-random amount of tabs it seems :D. If you can undo those changes...

I also see you have a `sub1_from_channel()` but in the can_join hook you should just return error (as you already do). In `can_join` in the core code (src/modules/join.c) there is no sub1_from_channel for other cases either, which implies this should not be done.

ValwareIRC:
I only sub1_from_channel() as a quick way to clean up the channel which, without it, would still exist in the LIST output with 0 users, and also be unjoinable.

syzop:
Ok, then that is an indication our code is wrong. Fixed now in https://github.com/unrealircd/unrealircd/commit/33c6eb0bcf6a10b3a62270214528cda41347577b. Thanks!

ValwareIRC:
And fixed! =]

PR #287, opened 2024-06-15
MODE: Reply with ERR_NOSUCHCHANNEL when the target is a channel

progval:
While it is common for IRC servers to use ERR_NOSUCHNICK instead of ERR_NOSUCHCHANNEL when a target can be either a channel or a nick, it seems every other IRCd but ngIRCd uses ERR_NOSUCHCHANNEL in this particular case. (similar PR sent to ngIRCd: https://github.com/ngircd/ngircd/pull/319)

### DIFF
--- a/src/modules/mode.c
+++ b/src/modules/mode.c
@@ -107,7 +107,7 @@ CMD_FUNC(cmd_mode)
 			channel = find_channel(parv[1]);
 			if (!channel)
 			{
-				CALL_CMD_FUNC(cmd_umode);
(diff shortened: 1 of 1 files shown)


ValwareIRC:
Indeed, according to the code, it's strange that it tried to convert to a umode command when the code already checked the target begins with a `#` which a nick cannot.

syzop:
Yeah, no biggy/violation i think, but indeed this makes no sense. Why would we call the umode handler for this case. No idea what happened there. Fixed now, thanks! :D.

PR #288, opened 2024-06-22
Verify upstream issuer's name can be used properly in places on IRC

ValwareIRC:
I made an account on my webpanel with a space in the username just to see what would happen, changed the topic of a channel and things got a bit messy due to explicitly trusting that it can be used on IRC.

I'm going to fix that in the webpanel also, but it doesn't stop other devs from potentially breaking things with their own software. So I think we should just do this to be safe.

### DIFF
--- a/src/modules/issued-by-tag.c
+++ b/src/modules/issued-by-tag.c
@@ -124,7 +124,14 @@ void _mtag_add_issued_by(MessageTag **mtags, Client *client, MessageTag *recv_mt
 		// TODO: test with all of: local rpc through unix socket, local rpc web, RRPC
 		if (client->rpc->issuer)
 		{
-			snprintf(buf, sizeof(buf), "RPC:%s@%s:%s", client->rpc->rpc_user, client->uplink->name, client->rpc->issuer);
+			// For several places in the source, like TOPIC
+			// we cannot explicitly trust the uplink issuer's
+			// name because it might contain spaces and that's
+			// not good. It must conform to IRC nick standards.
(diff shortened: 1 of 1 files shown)


syzop:
I see. While the code may look logical and correct at first sight, the `do_nick_name(client->rpc->issuer)` will cause the client->rpc->issuer to be altered. And changing client->rpc->issuer from issued-by-tag is rather unexpected (and bad).

We should either:
1) Sanitize and reject the issuer where the RPC call gets done
And/or
2) Sanitize it in issued-by-tag (not the client->rpc->issuer but use a temporary variable so to not touch the original).

But, I think I should first look at the underlying problem how big it is. You mention causing a mess but didn't really say what happened exactly. Looking quickly at the code I think it is more that the set_by value is unchecked while it should be (but that is technically unrelated to issuer). Anyway, working on other stuff atm.

ValwareIRC:
For example, when my webpanel username is "test username with spaces", and I go to set a topic to "lol", it looks like this:
```-> irc.valware.uk TOPIC #test
<- @time=2024-07-11T18:44:22.193Z :irc.valware.uk 332 Valware #test :with spaces 1720723428 :lol
<- @time=2024-07-11T18:44:22.193Z :irc.valware.uk 333 Valware #test test 1720723428
```
which gets rendered like this: `[02:43:48] * irc2.valware.co.uk changes topic to 'with spaces 1720723428 :lol'`

syzop:
Moved to https://bugs.unrealircd.org/view.php?id=6432

PR #289, opened 2024-07-11
crule: add "match_operclass(x)"

ValwareIRC:
Allows creating security-groups matching operclasses, useful for grouping your opers


ValwareIRC:
Closed because it was added by Syzop after I made this PR

PR #290, opened 2024-07-14
underscore is not special in simple match

crazycatdevs:
Remove the exception on underscore in match_simple


syzop:
Seems our messages crossed. I am handling this in https://bugs.unrealircd.org/view.php?id=6431. I already had a patch but "parked it". Closing this one since it is redundant then.

PR #292, opened 2024-07-18
Update crule.c: Add `idle_time()`

ValwareIRC:
useful for identifying active or inactive users


ValwareIRC:
Bump, this would be really handy for me

syzop:
This is now implemented as part of the bunch of crule functions that were added in https://github.com/unrealircd/unrealircd/commit/ed16dad40e9ff5c1ce5873e805f5d17450a07edc

PR #299, opened 2024-09-16
rpc/log: Add ability to send log messages via RPC

ValwareIRC:


ValwareIRC:
sorry I think I got a bit ahead of myself here, brb with some error checking

PR #300, opened 2024-09-16
sasl.c: Fix an underflow issue in `decode_authenticate_plain()`

ValwareIRC:


syzop:
Good find. Traced the source of the problem to `b64_decode()` and fixed it in 8e8384628b3d4ac00f6dd39fc44e3907d2e70022.

PR #301, opened 2024-09-23
wocket.c - Update buffer size to account for messagetags

ValwareIRC:
Allow full mtag messages to be sent over wockets
Special thanks to alice for pointing out the buffer length issue and specifying the message-tag length


syzop:
I don't understand how nobody could notice this, even with usual message tags like server-time and id's you loose dozens of characters. The code has been this way for 3 years...

syzop:
Did a follow-up commit in 7a43448674fabae1672b8e7ef6bc6e858c59f9d8 to fix some corner-case-issues.. like many replacement chars causing >510 bytes (excluding message tags).

PR #304, opened 2024-10-14
tkl.c: Update param checking

ValwareIRC:
Now this works like:
if the time param exists, even without a reason, it will be checked if it's a time param. if it's not a time param, it'll be considered to be the reason (or the first part of it anyway)


syzop:
"GLINE \*@1.2.3.4 0 test" was broken by this, or.. well.. it ignored the reason:
G-Line added: '\*@1.2.3.4' [reason: no reason] ... [duration: permanent]

Also has memory leaks due to not freeing 'reason'. It frees it at the end but there are two return conditions prior that where it isn't freed. Using the stack is a better idea, then we don't have this potential for alloc+free mistakes.

syzop:
Fixed these issues in 985a591df2f8cc44698a11f7189c63afbf3193c6. Hopefully correct now.

PR #305, opened 2024-10-21
Update help.conf

TehPeGaSuS:
Extended GLINE help with an example how to to remove a GLINE from a specific IP because, sometimes, people need exact syntaxes. :grin:


syzop:
I think that's a little bit over the top ;D.

Did add other enhancements though in 1f3c9b1dd29e7e6a972b4cf1c2573f4c3f5239ab

PR #308, opened 2025-02-06
restrict-commands: Allow messaging with ircops & help channel

ValwareIRC:
This allows users to send private-messages to ircops still, despite any config restricting privmsgs.
Similarly, also allow messages to the help channel.

### DIFF
--- a/src/modules/restrict-commands.c
+++ b/src/modules/restrict-commands.c
@@ -357,9 +357,15 @@ int rcmd_block_message(Client *client, const char *destination, const char *text
 	crule_context context;
 	RestrictedCommand *rcmd;
 	static char errbuf[256];
+	Client *target;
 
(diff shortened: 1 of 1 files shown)


syzop:
I don't think I would want this by default. I mean, we don't bypass spamfilter either if target is an IRCOp. Could be quite annoying for them.

The help channel exemption is already possible see https://www.unrealircd.org/docs/Restrict_commands#Restrict_messaging:
```
set {
        restrict-commands {
                channel-message {
                        except {
                                connect-time 600;
                                destination "#help";
                        }
                }
         }
}
```

We could add something to mask items to have like a destination-ircop or something else. Things to think about. I'm closing this one and discussing it further in https://bugs.unrealircd.org/view.php?id=6496

PR #309, opened 2025-02-24
Update example.pt.conf to the latest changes

devnull-hub-lab:
Update example.pt.conf to the latest changes, from April/2024 until now.


syzop:
Sorry for the delay :D

PR #310, opened 2025-04-29
Implement IRCv3's draft/extended-isupport

ValwareIRC:
https://github.com/ircv3/ircv3-specifications/blob/master/extensions/extended-isupport.md

### DIFF
--- a/src/modules/nick.c
+++ b/src/modules/nick.c
@@ -806,8 +806,11 @@ void welcome_user(Client *client, TKL *viruschan_tkl)
 	sendnumeric(client, RPL_MYINFO, me.name, version, umodestring, cmodestring);
 
 	RunHook(HOOKTYPE_WELCOME, client, 4);
-	for (i = 0; ISupportStrings[i]; i++)
-		sendnumeric(client, RPL_ISUPPORT, ISupportStrings[i]);
+
+	/* If user has draft/extended-isupport, assume they already got ISUPPORT tokens */
+	if (!HasCapability(client, "draft/extended-isupport"))
+		for (i = 0; ISupportStrings[i]; i++)
+			sendnumeric(client, RPL_ISUPPORT, ISupportStrings[i]);
 
 	RunHook(HOOKTYPE_WELCOME, client, 5);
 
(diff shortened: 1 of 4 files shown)


src/modules/nick.c
```diff
@@ -806,8 +806,11 @@ void welcome_user(Client *client, TKL *viruschan_tkl)
 	sendnumeric(client, RPL_MYINFO, me.name, version, umodestring, cmodestring);
 
 	RunHook(HOOKTYPE_WELCOME, client, 4);
-	for (i = 0; ISupportStrings[i]; i++)
-		sendnumeric(client, RPL_ISUPPORT, ISupportStrings[i]);
+
+	/* If user has draft/extended-isupport, assume they already got ISUPPORT tokens */
```
alicetries, line 810: This isn't quite a correct assumption based on the letter of the [spec](https://github.com/ircv3/ircv3-specifications/blob/master/extensions/extended-isupport.md?plain=1#L52-L53).
> The server MAY skip the `RPL_ISUPPORT` replies usually sent when connection registration completes, ***if it already sent all information***.

Only if the CAP has been negotiated and the user has actually sent the ISUPPORT command would it be valid for us to skip sending the RPL_ISUPPORT messages.

src/serv.c
```diff
@@ -188,14 +188,13 @@ char *getosname(void)
...
-		if (remote)
+	if (remote)
+		for (i = 0; ISupportStrings[i]; i++)
 			sendnumeric(client, RPL_REMOTEISUPPORT, ISupportStrings[i]);
-		else
-			sendnumeric(client, RPL_ISUPPORT, ISupportStrings[i]);
-	}
+
+	else
+		do_cmd(client, NULL, "ISUPPORT", 2, NULL);
```
alicetries, line 196: Would it be better to keep the code for sending ISUPPORT in here, and calling into it from extended-isupport? 

Thinking from the perspective of a server admin who has disallowed loading of the extended-isupport module here.

syzop:
Implemented in e78a6a6dbf625593c979c446c7e6c1f4d217bf36, 5cb2428567b1da03b654b2438947063614b7a2aa and 595f56007b1e1829850dc4ba0f7afb29ae40bff6. Tried to separate the initial merge-of-some-stuff vs the code cleanups... but.. yeah.

If you could test it, that would be appreciated :)

Closing the issue speculatively.

PR #311, opened 2025-06-10
Tell clients about `CHANMODE`, `PREFIX` and `STATUSMSG` changes at runtime

ValwareIRC:
This re-sends these ISUPPORT tokens to let users know about important changes that may affect their display, most notably `PREFIX` can be problematic; for example if you have a server running and wish to load a module like ojoin or something else that relies on the client knowing the correlation between the mode and the prefix char, and without it the client just doesn't display the nicklist properly from then on until the client reconnects, which as we know can be a while until that happens.

The expected client reaction to duplicate ISUPPORT tokens according to the spec is to overwrite the current values. I have tested this in mIRC only and it works as expected.


syzop:
Further commit in ba8c587e44d3e42035f9ac2dbd82bfcb6e1347d9 and discussion in https://bugs.unrealircd.org/view.php?id=6537

PR #312, opened 2025-07-05
Add `HOOKTYPE_CAN_USE_NICK`, useful for blocking on an internal level like "IRC" nick

ValwareIRC:
Useful for my use-case with Runa, this lets modules block nick changes on an internal level. This is suitable for module-provided bots (like Runa) where this idea is better than the module adding a qline for the nick, which can be subsequently removed by an oper (using services or third-party mods) and potentially cause issues by `/nick`ing to the "internal" nick.

### DIFF
--- a/include/modules.h
+++ b/include/modules.h
@@ -1542,7 +1542,7 @@ int hooktype_pre_chanmsg(Client *client, Channel *channel, MessageTag **mtags, c
  * @retval HOOK_DENY		Deny the message. The 'errmsg' will be sent to the user.
  * @retval HOOK_CONTINUE	Allow the message, unless other modules block it.
  */
-int hooktype_can_send_to_user(Client *client, Client *target, const char **text, const char **errmsg, SendType sendtype);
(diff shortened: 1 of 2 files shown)


include/modules.h
```diff
@@ -1526,7 +1524,7 @@ int hooktype_local_kick(Client *client, Client *victim, Channel *channel, Messag
 int hooktype_remote_kick(Client *client, Client *victim, Channel *channel, MessageTag *mtags, const char *comment);
 
 /** Called right before a message is sent to the channel (function prototype for HOOKTYPE_PRE_CHANMSG).
- * This function is only used by delayjoin. It cannot block a message. See hooktype_can_send_to_channel() instead!
+ * This function is only used by delayjoin. It cannot block a message. See hooktype_can_send_to_user() instead!
```
ValwareIRC, line 1527: This is a typo fix unrelated to the PR, but fixes a link in the doxygen pages

PR #315, opened 2025-07-10
Update modules.optional.conf

TehPeGaSuS:
Add `websocket` to the antirandom except list, since it will often trigger it because, by default (at least with KiwiIRC), the user ident will be the same as their nickname.


syzop:
I don't think that's a good idea. We could add some comment about KiwiIRC in the config file about this though, because - like you said - that situation is pretty common and is often fixed by just adjusting some kiwi config thingy. Let's talk about that on IRC.

PR #316, opened 2025-07-31
Adjust spamfilter:input-conversion to support 'deconfuse' as a synonym for 'confusables'

alicetries:
Adjust input-conversion to support 'deconfuse' as a synonym for 'confusables', as the release notes imply that 'deconfuse' should be valid for spamfilter:input-conversion.


syzop:
Good point. Although the current code uses `confusables`, the release notes and the documentation at https://www.unrealircd.org/docs/Spamfilter_block use `deconfused`, and your commit adds `deconfuse`. What a confusion!

PR #317, opened 2025-09-11
Fix issuer problem (doesn't show) in RPC TKLs

ValwareIRC:
This was reported in #unreal-webpanel by @TehPeGaSuS 

Expected behavior was to show the issuer in places like this, but wasn't being shown.

This adds in the proper issuer when none is explicitly mentioned in the call, falling back to the `client->name`


ValwareIRC:
Closes https://bugs.unrealircd.org/view.php?id=6574

PR #318, opened 2025-09-11
Add webhook.c

ValwareIRC:
This is useful since it allows us to push to a web application without the web application maintaining a persistent connection through RPC 'log' subscription. This will be useful for things like the webpanel because then extensions can be developed that can keep track of things and extend functionality (like sending out email alerts or whatever) without a user needing to be logged-into and present for the webpanel to get important notifications.


syzop:
As discussed on IRC: this is a nice idea, but probably better to integrate it in the log { } block as a destination webhook.

ValwareIRC:
I've royally messed up that PR, so I opened another: #322

PR #319, opened 2025-09-11
Include away status and reason in user JSON

ValwareIRC:
Add away information to user JSON object.

Closes https://bugs.unrealircd.org/view.php?id=6573


syzop:
Thanks. I went for a non-child object myself, and only show the 2 fields if the user is away, other than that it is exactly the same code. Done in https://github.com/unrealircd/unrealircd/commit/f42bab778e3bd6096853db651e0683f5d8d8ccea / will be in future UnrealIRCd 6.2.1.

PR #320, opened 2025-09-30
Fix multi-prefix

progval:
Broken by 15c8b1aa40be2cc65db292ccffd455303f51af07

### DIFF
--- a/src/modules/whox.c
+++ b/src/modules/whox.c
@@ -124,7 +124,7 @@ MOD_INIT()
 
 MOD_LOAD()
 {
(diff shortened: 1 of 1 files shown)


syzop:
Ah yeah, I made a similar mistake earlier in 792eca7d4d220963636dc20f1bfd7cd6349499e3. Have now scanned the complete 2-day-diff and none of these should be left. Also added a WHO test to our multi-prefix test (it was only testing NAMES before).

Thanks for the PR, merged :D

PR #323, opened 2025-10-30
message-tags: Fix length checks

progval:
The fix in da703efdf44d2599233aad9445fd00666427eda5 was incorrect because it swapped the limits for clients and dservers


progval:
Ugh of course this doesn't compile. Give me a sec.

syzop:
Thanks! :)

PR #329, opened 2025-12-04
Add help documentation for MONITOR command

TehPeGaSuS:


ValwareIRC:
Just wanna point out that MONITOR is a client command, not really intended for use by the user (much like PING or CAP), and [so the documentation](https://ircv3.net/specs/extensions/monitor) should really be the only documentation for the command. If you're having to type this command then you'll also be having to try and figure out what "731" means and why is IRC saying that? Because the response it also meant for the client to display in a cute way to the user.

syzop:
Yeah it's a bit mixed now, the `USERCMDS`, isn't it. I remember the point of view that we should not document commands that are not to be used by end-users. So then why does it include `WATCH` and `ISON`. I guess `PING` kinda makes sense because it allows you to ping other servers as well, `PONG` doesn't quite fit but i guess it comes with the `PING`. It's not really consistent. 

I'm leaning to removing a few of these to make it really for users only. For commands that should not be used by end-users and are not unrealircd specific they should use docs like RFC1459 and the horse docs thing.

syzop:
In commit 2868c3fedb0f62c648dc31045f84d94d851b47a5 I have tried to be more consistent with documenting only end-user-commands (scratching WATCH and ISON for instance) and added the horse docs reference for client developers.

I think this can be closed now.

PR #336, opened 2026-04-26
reply-tag, no-implicit-names: support ratified versions

ValwareIRC:
The IRCv3 specifications for these have been ratified:
- https://ircv3.net/specs/client-tags/reply
- https://ircv3.net/specs/extensions/no-implicit-names

Both the draft and ratified names are now supported for the transition period. Neither draft form is removed.

reply-tag.c:
- Register +reply handler (ratified) alongside +draft/reply
- Both carry MTAG_HANDLER_FLAGS_FIRST_ONLY so neither propagates to subsequent lines of a multiline message
- mtag_add_replytag propagates whichever form the client sent

no-implicit-names.c:
- Register no-implicit-names CAP (ratified) alongside draft/no-implicit-names
- Both are active; clients may negotiate either

join.c:
- _join_channel now suppresses the implicit NAMES reply when the client has negotiated either draft/no-implicit-names or no-implicit-names

### DIFF
--- a/src/misc.c
+++ b/src/misc.c
@@ -1361,7 +1361,7 @@ MessageTag *duplicate_mtags(MessageTag *mtags)
 
 /** Duplicate a message tag list, excluding tags with MTAG_HANDLER_FLAGS_FIRST_ONLY.
  * Used to build the tag set for subsequent lines (lines 2..N),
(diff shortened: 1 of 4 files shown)


syzop:
Will look at this for 6.2.6.

PR #337, opened 2026-04-27
chathistory: implement draft/chathistory-end tag

ValwareIRC:
This is a recent [spec modification suggestion](https://github.com/ircv3/ircv3-specifications/pull/598) which I agree is a nice touch, I'm not leaving this as a draft like the [other PR](https://github.com/unrealircd/unrealircd/pull/334) because I think this one will have virtually no impact whatsoever if implemented immediately.

This implements a new tag on a batch when the batch is the final batch of a chathistory request, as in "this batch contains the final message in the chathistory so there are no more messages to send, so you can stop requesting more/hide UI button for viewing more messages"

### DIFF
--- a/include/h.h
+++ b/include/h.h
@@ -1218,7 +1218,7 @@ extern HistoryResult *history_request(const char *object, HistoryFilter *filter)
 extern int history_delete(const char *object, HistoryFilter *filter, int *rejected_deletes);
 extern int history_destroy(const char *object);
 extern int can_receive_history(Client *client);
-extern void history_send_result(Client *client, HistoryResult *r);
+extern void history_send_result(Client *client, HistoryResult *r, int end_of_pagination);
 extern void free_history_result(HistoryResult *r);
 extern void free_history_filter(HistoryFilter *f);
 extern void special_delayed_unloading(void);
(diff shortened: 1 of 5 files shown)


syzop:
This looks useful to clients and i see it was [merged](https://github.com/ircv3/ircv3-specifications/commit/9f65105eda718666dbc6cf4409d2c68f2dd9c73f) in the draft/chathistory spec on May 14 2026. Will have a look at this for 6.2.6.

syzop:
Thanks for the PR.

I did a couple of follow-up fixes: https://github.com/unrealircd/unrealircd/commit/37977fcfe62755267c8fcafc5624b6f0727d3eed and https://github.com/unrealircd/unrealircd/commit/570c32ea6705ccf3f9907fa67c78320707e1e8cb (the latter is unrelated to your code, is an existing issue)

I also noticed we don't always send an end tag when we actually reach the end. That's safe.. the client would just make another request to get next stuff and it would return 0 results (just like.. current situation, basically), but is a bit of a missed opportunity. It's a more complex case of where the returned count is the same as the limit (count==limit), so it moves the problem to another layer (not chathistory module). I'll see if that can be improved... no need for a new PR for that from anyone.

And again, thanks, this sounds like useful stuff for clients.

syzop:
I had to redo this feature because of.. see comment above.. but it is commited now: https://github.com/unrealircd/unrealircd/commit/09a732e2c146836972a24eb34e4934d6803071b7

PR #346, opened 2026-09-13
Fix STATUSMSG delivery reaching ranks below the intended target

TehPeGaSuS:
## Summary

`channel_member_modes_generate_equal_or_greater()` in `src/api-channelmode.c` compares a numeric channel-mode rank (owner=4000, admin=3000, op=2000, halfop=1000, voice=-1) against the raw ASCII value of a mode letter returned by `lowest_ranking_mode()`, instead of converting it through `mode_to_rank()` first:

```c
rank = lowest_ranking_mode(modes);   /* returns a mode LETTER, e.g. 'q' = 113 */
if (!rank)
    return;

for (cm = channelmodes; cm; cm = cm->next)
    if ((cm->type == CMODE_MEMBER) && (cm->rank >= rank))   /* comparing rank vs ASCII code */
        strlcat_letter(buf, cm->letter, buflen);
```

Since every mode letter's ASCII code (`'v'`=118, `'h'`=104, `'o'`=111, `'a'`=97, `'q'`=113) is well under 1000, `cm->rank >= rank` is satisfied by halfop/op/admin/owner (ranks 1000-4000) **regardless of the actual requested target**. The only exception is voice (rank `-1`), which is incorrectly *excluded* even when it should match.

### Impact

- `NOTICE ~#channel` (intended for `+q`/owner only) is also delivered to `+o` and `+h` members.
- `NOTICE @#channel` (intended for `+o` and up) is also delivered to `+h` members.
- `NOTICE +#channel` (intended for `+v` and up) incorrectly *excludes* actual `+v` users, while still including `+h` and up.

This affects `PRIVMSG`/`NOTICE` STATUSMSG targeting (`@#channel`, `%#channel`, `~#channel`, etc.) network-wide on any UnrealIRCd 6 install using it - a real information-disclosure bug, not cosmetic.

### Fix

Convert the mode letter to its numeric rank via the existing `mode_to_rank()` helper before comparing:

```c
rank = mode_to_rank(lowest_ranking_mode(modes));
```

## Test plan

- [x] Reproduced on a live test instance: `NOTICE ~#channel :message` was visible to `@` (op) users, who should not have received it.
- [x] Read the delivery path end-to-end (`message.c`'s `cmd_message()` -> `send.c`'s `sendto_channel()` -> `channel_member_modes_generate_equal_or_greater()`) to confirm this is the only place the rank/letter confusion occurs.
- [x] Verified the fix with a full clean rebuild from a fresh scratch checkout (`./configure && gmake`) - zero compiler warnings/errors, `ircd` links successfully.


syzop:
Thanks! This was indeed the correct fix. Commited as https://github.com/unrealircd/unrealircd/commit/f9a64a8878dd3ff554f02d09e27936af3580bf6b.

This also fixes https://bugs.unrealircd.org/view.php?id=6491

